PR #1, opened 2020-04-30
Introduce default ASP.NET Core Web API template

ethax-ross:
Add the default template for an ASP.NET Core Web API project, with an example
resource that can be accessed at [https://localhost:5001/example](https://localhost:5001/example).


ethax-ross:
👍 Squashed in those changes @DrLazer

PR #3, opened 2020-05-04
Stub GiT controller actions, request models/validation

ethax-ross:
Stubs out the controller actions we will need for the Get into Teaching website (TTA registration, events and mailing list). The request models have been created where possible, but in places that will hook into the generated OData entities for Dynamics I've left the models as `Object`.

Model validation uses [FluentValidation](https://fluentvalidation.net/) and I've also added [FluentAssertions](https://fluentassertions.com/) and [Moq](https://github.com/Moq/moq4/wiki/Quickstart) for testing.

Started migrating the Swagger documentation into the source code using [Swashbuckle](https://github.com/domaindrivendev/Swashbuckle.AspNetCore) (which can [integrate nicely with FluentValidation](https://github.com/micro-elements/MicroElements.Swashbuckle.FluentValidation)) - when this has been fully migrated over we can remove `docs/swagger.yml`:

Screenshot 2020-05-04 at 09 39 23

### DIFF
--- a/GetIntoTeachingApi/Controllers/MailingListController.cs
+++ b/GetIntoTeachingApi/Controllers/MailingListController.cs
@@ -1,0 +1,41 @@
+﻿using GetIntoTeachingApi.Models;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+using Swashbuckle.AspNetCore.Annotations;
+using System;
+using System.Collections.Generic;
+
+namespace GetIntoTeachingApi.Controllers
+{
+    [Route("api/mailing_list")]
+    [ApiController]
+    public class MailingListController : ControllerBase
+    {
+        private readonly ILogger<MailingListController> _logger;
+
+        public MailingListController(ILogger<MailingListController> logger)
+        {
+            _logger = logger;
+        }
+
+        [HttpPost]
+        [Route("members")]
+        [SwaggerOperation(
+            Summary = "Adds a new member to the mailing list.",
+            Description = "Adds a new member to the mailing list. A new candidate will also be created if a matching candidate can not be found.",
+            OperationId = "AddMailingListMember",
+            Tags = new[] { "Mailing List" }
+        )]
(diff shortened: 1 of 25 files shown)


GetIntoTeachingApi/Controllers/MailingListController.cs
```diff
@@ -0,0 +1,41 @@
+﻿using GetIntoTeachingApi.Models;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+using Swashbuckle.AspNetCore.Annotations;
+using System;
+using System.Collections.Generic;
+
+namespace GetIntoTeachingApi.Controllers
+{
+    [Route("api/mailing_list")]
```
DrLazer, line 10: Conventionally the 'api/' part isn't added to each controller.  But it can be sorted later

GetIntoTeachingApi/Controllers/TeachingEventsController.cs
```diff
@@ -0,0 +1,104 @@
...
+            Tags = new[] { "Teaching Events" }
+        )]
+        public IActionResult Get([FromRoute, SwaggerParameter("The `readableEventId` of the `TeachingEvent`.", Required = true)] string readableEventId)
+        {
+            // TODO:
+            return Ok(new Object());
+        }
+
+        [HttpPost]
+        [Route("")]
```
DrLazer, line 69: Do you need this empty route pre-compiler directive?

ethax-ross: Nope 👍

GetIntoTeachingApi/Controllers/MailingListController.cs
ethax-ross: I was going by this (though I now notice they use `RoutePrefix` and not `Route`, but I don't think there's a difference?:

https://docs.microsoft.com/en-us/aspnet/web-api/overview/web-api-routing-and-actions/attribute-routing-in-web-api-2#route-prefixes

Is there somewhere else I can put a route prefix for all the controllers?

PR #4, opened 2020-05-05
Add Timed One Time Password (TOTP) for candidate access tokens

ethax-ross:
Adds a `CandidateAccessTokenService` class that is able to generate Timed One Time Password (TOTP) codes. These can then be emailed out and later verified (the verification also checks a shared secret that should be sent from the client consuming the API). 

Currently, the TOTPs are configured with a 30 second 'step', which means a new code will be generated every 30 seconds and they have a window size of 2, meaning the user will have 1 minute (plus the time left in the current 'step') in order to send the code back in a candidate request. We can change this later on if we want to allow longer (by adjusting the windows size - the step size should remain as per the RFC).

We configure the TOTP secret to be a compound value of candidate email address and shared secret; this results in different email addresses generating different tokens within the same step (time period), as the key is used in conjunction with the current time when deriving the token:

```
TOTP(K,T) = Truncate(HMAC-SHA-1(K,T))
```

If we didn't do this, then we would be vulnerable to a candidate generating a token for their own email address _and_ another candidate's email address -- the token would be valid/the same for both. As per the RFC the HMAC is truncated, which means that two different emails _could_ result in the same token, but chances of this should be very slim (and the tokens themselves are only 'live' for a handful of minutes which means it should be even slimmer).

### DIFF
--- a/GetIntoTeachingApi/Controllers/TeacherTrainingAdviser/CandidatesController.cs
+++ b/GetIntoTeachingApi/Controllers/TeacherTrainingAdviser/CandidatesController.cs
@@ -3,6 +3,8 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 using Swashbuckle.AspNetCore.Annotations;
 using System.Collections.Generic;
+using GetIntoTeachingApi.Services;
+using GetIntoTeachingApi.Models;
 
 namespace GetIntoTeachingApi.Controllers.TeacherTrainingAdviser
 {
@@ -11,10 +13,12 @@ namespace GetIntoTeachingApi.Controllers.TeacherTrainingAdviser
     public class CandidatesController : ControllerBase
     {
         private readonly ILogger<CandidatesController> _logger;
+        private readonly ICandidateAccessTokenService _tokenService;
 
-        public CandidatesController(ILogger<CandidatesController> logger)
+        public CandidatesController(ILogger<CandidatesController> logger, ICandidateAccessTokenService tokenService)
         {
             _logger = logger;
+            _tokenService = tokenService;
         }
 
         [HttpPost]
@@ -40,11 +44,18 @@ Retrieves an existing candidate for the Teacher Training Adviser service. The `a
             OperationId = "GetExistingTeacherTrainingAdviserCandidate",
             Tags = new[] { "Teacher Training Adviser" }
         )]
+        [ProducesResponseType(401)]
         public IActionResult Get(
-            [FromRoute, SwaggerParameter("Access token (PIN code).", Required = true)] string accessToken, 
-            [FromHeader(Name = "Authorization"), SwaggerParameter("Bearer <sharedSecret>.", Required = true)] string sharedSecret
+            [FromRoute, SwaggerParameter("Access token (PIN code).", Required = true)] string accessToken,
+            [FromHeader(Name = "Candidate-Email"), SwaggerParameter("Candidate email address.", Required = true)] string email
         )
         {
+            var challenge = new CandidateAccessTokenChallenge { Token = accessToken, Email = email };
+            if (!_tokenService.IsValid(challenge))
+            {
+                return Unauthorized();
+            }
+
             // TODO:
             return Ok(new Object());
         }
--- a/GetIntoTeachingApi/Services/CandidateAccessTokenService.cs
+++ b/GetIntoTeachingApi/Services/CandidateAccessTokenService.cs
@@ -1,0 +1,60 @@
+﻿using GetIntoTeachingApi.Models;
+using OtpNet;
+using System;
+using System.Text;
+
+namespace GetIntoTeachingApi.Services
+{
+    public class CandidateAccessTokenService : ICandidateAccessTokenService
+    {
+        // The amount of time a user has to verify their access token is:
+        // (VerificationWindow * StepsInSeconds) + Remaining Seconds in Current Step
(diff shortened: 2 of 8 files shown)


.vscode/launch.json
```diff
@@ -1,31 +0,0 @@
-{
```
jebw, line 1: Not a problem but personally I do drive by fixes like these to be off on their own PR?

GetIntoTeachingApi/Controllers/TeacherTrainingAdviser/CandidatesController.cs
```diff
@@ -40,11 +44,19 @@ public CandidatesController(ILogger<CandidatesController> logger)
...
         )]
+        [ProducesResponseType(401)]
         public IActionResult Get(
-            [FromRoute, SwaggerParameter("Access token (PIN code).", Required = true)] string accessToken, 
+            [FromRoute, SwaggerParameter("Access token (PIN code).", Required = true)] string accessToken,
+            [FromHeader(Name = "Candidate-Email"), SwaggerParameter("Candidate email address.", Required = true)] string email,
             [FromHeader(Name = "Authorization"), SwaggerParameter("Bearer <sharedSecret>.", Required = true)] string sharedSecret
         )
         {
+            var challenge = new CandidateAccessTokenChallenge { Token = accessToken, SharedSecret = sharedSecret, Email = email };
```
jebw, line 54: Bear with me, I could be reading this wrong (first time I've really looked at C#), _but_ this looks wrong - there's should be two secrets here

1. Bearer token in header used for server <--> client auth - _ideally we can support multiple for different clients, comma separated maybe?_
2. Private internal secret only available to server, this gets combined with email for generating TOTP

At present the API client has all the information to generate valid TOTP codes itself?

GetIntoTeachingApi/Controllers/TeacherTrainingAdviser/CandidatesController.cs
```diff
@@ -11,10 +13,12 @@ namespace GetIntoTeachingApi.Controllers.TeacherTrainingAdviser
     public class CandidatesController : ControllerBase
     {
         private readonly ILogger<CandidatesController> _logger;
+        private readonly ICandidateAccessTokenService _tokenService;
 
-        public CandidatesController(ILogger<CandidatesController> logger)
+        public CandidatesController(ILogger<CandidatesController> logger, ICandidateAccessTokenService tokenSerivce)
```
jebw, line 18: minor typo, `s/tokenSerivce/tokenService/`

GetIntoTeachingApi/Controllers/TeacherTrainingAdviser/CandidatesController.cs
```diff
@@ -11,10 +13,12 @@ namespace GetIntoTeachingApi.Controllers.TeacherTrainingAdviser
     public class CandidatesController : ControllerBase
     {
         private readonly ILogger<CandidatesController> _logger;
+        private readonly ICandidateAccessTokenService _tokenService;
 
-        public CandidatesController(ILogger<CandidatesController> logger)
+        public CandidatesController(ILogger<CandidatesController> logger, ICandidateAccessTokenService tokenSerivce)
         {
             _logger = logger;
+            _tokenService = tokenSerivce;
```
jebw, line 21: minor typo, `s/tokenSerivce/tokenService/`

GetIntoTeachingApi/Services/CandidateAccessTokenService.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        private Totp CreateTotp(string email)
+        {
+            return new Totp(CompoundSharedSecretBytes(email), totpSize: Length, step: StepInSeconds);
+        }
+
+        private byte[] CompoundSharedSecretBytes(string email)
+        {
+            return Encoding.ASCII.GetBytes(email + SharedSecret());
```
jebw, line 52: You mentioned that this value gets truncated, do you know if its hashed prior to truncation - or is it the resultant string thats truncated?

If its the compound string I'd suggest hashing this. Either SHA1 or SHA2-256, probably the latter if its going to get truncated - I don't think the cost of hashing will be a consideration.

```
return Encoding.ASCII.GetBytes(HypotheticalSHA2Function(email + SharedSecret())) ;
```

.vscode/launch.json
ethax-ross: Yeah that's fair, I'll pull it out

ethax-ross: https://github.com/DFE-Digital/get-into-teaching-api/pull/5

GetIntoTeachingApi/Services/CandidateAccessTokenService.cs
ethax-ross: The key gets hashed with the current time prior to truncation - the function is:

```
TOTP(K,T) = Truncate(HMAC-SHA-1(K,T))
```

GetIntoTeachingApi/Controllers/TeacherTrainingAdviser/CandidatesController.cs
ethax-ross: Yeah you're right in that I've used the same token for the shared secret and the TOTP secret; I didn't think the client being able to generate valid codes would be an issue as I believe this is the case with the TOTP algorithm anyway - but you make a good point that we're likely going to have multiple clients in the future so that should be separated 👍

PR #6, opened 2020-05-05
Send PIN code email via GOV.UK Notify service

ethax-ross:
Includes the GOV.UK Notify C# client and wraps it in an adapter so that its mockable in unit tests. This is utilised by a basic `NotifyService` in order to send emails using a given `templateId` and `personalisation` (template attributes).

The `CandidatesController` has been updated to generate and email out a PIN code to a candidate for a valid `CandidateAccessTokenRequest`. There is a `// TODO:` in the action which will later be the call out to the CRM to find a matching candidate (so we only send a PIN code if there is a matching candidate in the CRM).

Screenshot 2020-05-05 at 19 48 41

### DIFF
--- a/GetIntoTeachingApi/Controllers/CandidatesController.cs
+++ b/GetIntoTeachingApi/Controllers/CandidatesController.cs
@@ -1,9 +1,11 @@
 ﻿using GetIntoTeachingApi.Models;
 using Microsoft.AspNetCore.Authorization;
+using GetIntoTeachingApi.Services;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 using Swashbuckle.AspNetCore.Annotations;
 using System.Collections.Generic;
+using System.Threading.Tasks;
 
 namespace GetIntoTeachingApi.Controllers
 {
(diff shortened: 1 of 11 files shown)


GetIntoTeachingApi/Controllers/CandidatesController.cs
```diff
@@ -36,6 +46,13 @@ public IActionResult CreateAccessToken([FromBody, SwaggerRequestBody("Candidate
                 return BadRequest(this.ModelState);
             }
 
+            if (true) // TODO:
```
DrLazer, line 49: is this // TODO still supposed to be here?

ethax-ross: Yeah:

> There is a // TODO: in the action which will later be the call out to the CRM to find a matching candidate (so we only send a PIN code if there is a matching candidate in the CRM).

GetIntoTeachingApi/Startup.cs
```diff
@@ -22,6 +24,10 @@ public Startup(IConfiguration configuration)
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
+            services.AddScoped<INotificationClientAdapter, NotificationClientAdapter>();
```
DrLazer, line 27: Worth a check here that AddScoped is really what you want.

Services, quite often are scoped as Singletons with AddSingleton to keep the same object alive across multiple requests.

It depends on the functionality within each individual service.

GetIntoTeachingApi/Controllers/CandidatesController.cs
DrLazer: cool

GetIntoTeachingApi/Startup.cs
ethax-ross: Yeah that's a good shout, these could be singletons - I'll get that updated 👍

PR #10, opened 2020-05-11
Match CandidateAccessTokenRequest with existing candidate

ethax-ross:
We need to be able to retrieve a candidate from the CRM by email address so that we can match it to an incoming `CandidateAccessTokenRequest`.

Add basic `Candidate` model with minimal attributes (to be fleshed out in a later PR where we return the `Candidate` from our API).

Add profile to map from Dynamics `Entity` to our `Candidate` model.

When a request for an existing candidate arrives we compare email address as well as first, last name and date of birth with what we have stored in the CRM. If the email matches in addition to 2/3 of the additional attributes then we go ahead and email a pin code to the candidate email address (they can then exchange this for the existing candidate details).

### DIFF
--- a/GetIntoTeachingApi/Services/CrmService.cs
+++ b/GetIntoTeachingApi/Services/CrmService.cs
@@ -45,6 +45,19 @@ namespace GetIntoTeachingApi.Services
                 .First();
         }
 
+        public async Task<Candidate> GetCandidate(string email)
+        {
+            return (await _context.CreateQuery(ConnectionString(), "contact"))
+                .Where((contact) =>
+                    // Will perform a case-insensitive comparison
+                    contact.GetAttributeValue<string>("emailaddress1") == email
+                )
+                .OrderByDescending((contact) => contact.GetAttributeValue<DateTime>("createdon"))
+                .Select((candidate) => _mapper.Map<Candidate>(candidate))
+                .FirstOrDefault();
+        }
+
+
         private string ConnectionString()
         {
             return $"AuthType=ClientSecret; url={InstanceUrl()}; ClientId={ClientId()}; ClientSecret={ClientSecret()}";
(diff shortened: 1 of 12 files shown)


GetIntoTeachingApi/Services/CrmService.cs
```diff
@@ -45,6 +45,18 @@ public async Task<PrivacyPolicy> GetLatestPrivacyPolicy()
                 .First();
         }
 
+        public async Task<Candidate> GetCandidate(string email)
+        {
+            return (await _context.CreateQuery(ConnectionString(), "contact"))
+                .Where((contact) =>
```
jebw, line 51: Can I just clarify - for this piece of code `.where` will generate the appropriate URL params for a query in dynamics, than `.select` is doing a client side 'reduce' to extract the first matching result?

Assuming the above is correct - two things I think are needed, haven't worked out if they are already there or not.
1. Some limit on the number of results returned
2. Need to retrieve the results by most-recent first

NB: Its out of scope of this piece of work but in SE we ended up tracking contacts we had matched before so we could favour those over new contacts added for the same email addresses.

ethax-ross: `Select` is a deferred operator; it will be executed once the results are loaded into memory by an immediate operator (`First` in this case). `First` also essentially equates to `Limit(1)` but returns a single value instead of a collection (I've maybe confused things by putting `Select` before `First`, but that seems to be the convention).

That's a good point about taking the most recent results; I had assumed that there would only be one contact with a particular email address in Dynamics. I'll add an ordering in by the created date 👍 we can maybe look at the last point at the end as a nice to have?

jebw: Conclusion from discussion was the logic of this is still being finalised so we'll go with this for now and iterate it later if needed once the matching rules have been clarified.

PR #24, opened 2020-05-19
Update README with architecture overview

ethax-ross:


ethax-ross:
Closing to capture this in an [ADR](https://github.com/joelparkerhenderson/architecture_decision_record) instead.

PR #25, opened 2020-05-19
Add in-memory cache for CrmService

ethax-ross:
The in-memory cache will store responses from the CRM that are used in validation logic, so that we can still accept requests from a client when the CRM is offline. This covers:

- Lookup items
- PickList items
- Privacy policy entities

The cache will be refreshed every 3 hours, however if the CRM is offline when a refresh is attempted the stale data will remain available in the cache so that the API can continue to function.

As this cache is in-memory, if the API server restarts when the CRM is offline it will not be able to function until the CRM comes back online. We expect the chance of this happening to be slim and will monitor our logs to see if it is worth the effort of implementing a persistent cache.

### DIFF
--- a/GetIntoTeachingApiTests/Services/CrmCacheTests.cs
+++ b/GetIntoTeachingApiTests/Services/CrmCacheTests.cs
@@ -1,0 +1,65 @@
+﻿using System;
+using FluentAssertions;
+using GetIntoTeachingApi.Services;
+using GetIntoTeachingApiTests.Utils;
+using Microsoft.Extensions.Logging;
+using Moq;
+using Xunit;
+
+namespace GetIntoTeachingApiTests.Services
+{
+    public class CrmCacheTests
+    {
+        private readonly Mock<ILogger<CrmCache>> _mockLogger;
+        private readonly CrmCache _cache;
+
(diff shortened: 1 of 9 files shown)


GetIntoTeachingApiTests/Services/CrmCacheTests.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        [Fact]
+        public void GetOrCreate_WhenRefreshFails_ReturnsStaleData()
+        {
+            _cache.GetOrCreate("key", DateTime.Now.AddSeconds(-30), () => "value");
+
+            var result = _cache.GetOrCreate("key", DateTime.Now.AddSeconds(30), () =>
+            { 
+                throw new Exception("bang");
+                return "";
```
DrLazer, line 58: this return is triggering some warnings

ethax-ross: Yeah this can be ignored, it's in the test suite and I need to test the exception being raised in a functional argument, don't think there's a way to do it without the warning but I'll take another look and pick it up in my next PR if there is

PR #26, opened 2020-05-20
Add upcoming events endpoint

ethax-ross:
- Add endpoint to retrieve teaching event types
- Add initial `TeachingEvent` model
- Add `UpcomingTeachingEvents` to `CrmService`
- Plumb in upcoming teaching events endpoint to `CrmService`

All events are currently loaded into the in-memory cache, however it is TBC if events need to be available whilst the CRM is offline.


ethax-ross:
Yeah its been irritating me for a while that model - ideally it would be called `Type` but that's reserved. It's not anything to do with events - it's a model that represents a generic 'type' returned from the CRM. Open to suggestions, but something I would pick up in a separate PR

PR #27, opened 2020-05-20
Suppress unreachable code warning

ethax-ross:
We are testing that a functional parameter throws an exception; the function requires a string return type but it will never be reached as we always throw the exception. This commit suppresses the build warning about unreachable code.


ethax-ross:
I don't think so - `GetOrCreate` takes a functional parameter that demands a return type - in this case it boils down to `Func<string> create` - if you don't return a string in that function you'll get a build error (or a different compiler warning)... if it was a suppression in the application I would agree that it needs a different solution, but this is just manufacturing a particular scenario for the test

PR #28, opened 2020-05-20
Add endpoint for searching events

ethax-ross:
- Add UK postcode lat/lng data set
- Add PostcodeService for validation/geocoding
- Add ability to search for events by radius around postcode, date and type

Introduces a second way of querying data using `QueryExpression` instead of `LINQ` as a limitation of the `LINQ` queries means that we can't return a collection and relationships in a single request. The side-effect of this is that related entities are loaded differently using `QueryExpression`; they get left-outer joined into the parent entity so we need to account for both methods when mapping the entities into our models. Once the API is feature complete I will spike moving from the SDK to the WebAPI as that seems to support both deep insert and querying related entities in a single request much better than the SDK.

Postcode data is stored in-memory; this causes the application to be slow when booting, so we will likely move this to a database in a future PR. In development a slim copy of the postcode csv is loaded for speed.

Best reviewed by-commit as adding in the large CSV of postcodes has caused GitHub to collapse everything in the `Files changed` tab.


GetIntoTeachingApi/Services/PostcodeService.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        private void LoadPostcodesFixture(string path)
+        {
+            _postcodes.Clear();
+
+            using var reader = new StreamReader(path);
+            using var csv = new CsvReader(reader, CultureInfo.InvariantCulture);
+
+            csv.Read();
```
DrLazer, line 49: does this file operation have its own error handling, or does this need a check for a problems occurred when reading?

GetIntoTeachingApi/Services/PostcodeService.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        public double? Latitude { get; set; }
+        public double? Longitude { get; set; }
+
+        public GeoCoordinate Coordinate
+        {
+            get { return IsNonGeographic() ? null : new GeoCoordinate((double) Latitude, (double) Longitude); }
+        }
+
+        public bool IsNonGeographic()
```
DrLazer, line 84: just a thought here, would IsGeographic() make the code read a bit nicer

GetIntoTeachingApi/Services/PostcodeService.cs
ethax-ross: I don't think so; there's no recovering if the CSV is dodgy, I would want it to raise an exception

GetIntoTeachingApi/Adapters/OrganizationServiceAdapter.cs
```diff
@@ -21,17 +22,28 @@ public IQueryable<Entity> CreateQuery(string entityName, OrganizationServiceCont
...
+        }
+
         public void LoadProperty(Entity entity, Relationship relationship, OrganizationServiceContext context)
         {
             context.LoadProperty(entity, relationship);
         }
 
         public IEnumerable<Entity> RelatedEntities(Entity entity, string attributeName)
         {
-            return entity.RelatedEntities
```
DrLazer, line 31: var result = new List<Entity>();
result.Add( <stuff> )
return result;

GetIntoTeachingApi/Services/PostcodeService.cs
ethax-ross: Yeah I thought about that but settled on `IsNonGeographic` as that seems to be the terminology used - you get `postcodes` and `non geographic postcodes` - I think it makes it clearer that some postcodes don't have geolocations (which I didn't know until I did this!) -- happy to change though if you disagree, I don't feel that strongly on it

GetIntoTeachingApi/Models/TeachingEventSearchRequest.cs
```diff
@@ -0,0 +1,37 @@
...
+        [SwaggerSchema("Set to filter results to a type of teaching event. Must match an `typeId` of the `TeachingEvent` schema.")]
+        public int? TypeId { get; set; }
+        [SwaggerSchema("Set to filter results to those that start after a given date.")]
+        public DateTime? StartAfter { get; set; }
+        [SwaggerSchema("Set to filter results to those that start before a given date.")]
+        public DateTime? StartBefore { get; set; }
+
+        public bool Match(TeachingEvent teachingEvent, IPostcodeService postcodeService)
+        {
+            if (TypeId != null && TypeId != teachingEvent.TypeId)
```
DrLazer, line 22: Omitting the { brackets is generally seen as bad practice.  Im sure resharper disagrees with me

ethax-ross: Hah yeah I'm afraid it does - I agree usually but since C# wants the brackets on a separate line it irks me 😄

PR #29, opened 2020-05-22
Plumb in endpoint to get single event

ethax-ross:
- Plumb in the endpoint to get a single event
- Replace `CandidateIdentification` with `ExistingCandidateRequest` model
- Drop create event endpoint


GetIntoTeachingApi/Services/CrmService.cs
```diff
@@ -76,6 +76,13 @@ public IEnumerable<TeachingEvent> SearchTeachingEvents(TeachingEventSearchReques
                 .OrderBy(teachingEvent => teachingEvent.StartAt);
         }
 
+        public TeachingEvent GetTeachingEvent(Guid id)
+        {
+            return GetTeachingEvents()
+                .Where((teachingEvent) => teachingEvent.Id == id)
+                .FirstOrDefault();
```
DrLazer: do you need the Where() ?
can it be done in the FirstOrDefaiult() ?

FirstOrDefault((teachingEvent) => teachingEvent.Id == id)

PR #42, opened 2020-05-29
Migrate postcode geolocation data into a database

ethax-ross:
The postcode information will be stored in Postgres going forward; it makes more sense for that model to be called `Location` instead of `Postcode` as one of the attributes will be `postcode`.

There are roughly 1.7 million UK postcodes that we currently read from a CSV file and store in memory to do geolocation comparisons when searching for events by radius.

This commit introduces Entity Framework, backed by Postgres and a `Location` model; the postcode and associated geolocation are read from the CSV and seeded into the database (if not yet present) when the application starts.

The `LocationService` has been updated to query postcode information from the database and the in-memory store has been removed.

Entity Framework is configured to use an in-memory SQLite database when in development.

Going forward it will make sense to also put event information into Postgres so that we can run a query to find events within a geographical region, instead of querying postcode information individually for each event as we currently do.

Also fixes a bug with radius-based event search when the event address is omitted and/or doesn't include a postcode.


GetIntoTeachingApi/Database/DbConfiguration.cs
```diff
@@ -0,0 +1,78 @@
...
+        public void Configure()
+        { 
+            _dbContext.Database.Migrate();
+
+            SeedLocations();
+        }
+
+        private void SeedLocations()
+        {
+            if (_dbContext.Locations.Any()) return;
```
DrLazer, line 32: stick some brackets around this old bean

PR #51, opened 2020-06-01
continue on error added

sfawcett123:
ok using continue on error, hopefully that will allow us to record the lack of artifact but continue,


sfawcett123:
really sorry there is no way to test these actions outside of github,

PR #70, opened 2020-06-05
Set application to stopped on deploy

sfawcett123:
The application will not start due to a bug, but for easy of deployment lets not try and start it automatically until we fix bugs


Dockerfile
```diff
@@ -3,6 +3,12 @@ FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS build
 WORKDIR /source
 ENV ASPNETCORE_URLS=http://+:8080
 
+# Install dependencies
+# hadolint ignore=DL3008
+RUN apt-get update && apt-get install -y --no-install-recommends  \
+	libsqlite3-mod-spatialite \
```
ethax-ross, line 9: We don't need `libsqlite3-mod-spatialite` in the Dockerfile any more, just for CI

PR #79, opened 2020-06-08
Delete orphaned models during sync

ethax-ross:
The entities and associated metadata in the CRM may be removed as well as added/updated; the sync operation needs to account for this by removing any models that are cached in the database but are no longer returned from the CRM.

### DIFF
--- a/GetIntoTeachingApi/Services/Store.cs
+++ b/GetIntoTeachingApi/Services/Store.cs
@@ -113,13 +113,13 @@ namespace GetIntoTeachingApi.Services
             var buildings = teachingEvents.Where(te => te.Building != null)
                 .Select(te => te.Building).DistinctBy(b => b.Id);
 
-            UpsertAndRemoveOrphanedModels(buildings, _dbContext.TeachingEventBuildings);
+            SyncModels(buildings, _dbContext.TeachingEventBuildings);
 
-            // Link events with buildings attached to the context prior to upsert.
+            // Link events with buildings attached to the context prior to sync.
             teachingEvents.Where(te => te.Building != null).ToList()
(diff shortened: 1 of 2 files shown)


GetIntoTeachingApi/Services/Store.cs
```diff
@@ -113,58 +113,64 @@ private void SyncTeachingEvents(ICrmService crm)
...
+            UpsertAndRemoveOrphanedTypes(crm.GetPickListItems("dfe_qualification", "dfe_degreestatus"));
+            UpsertAndRemoveOrphanedTypes(crm.GetPickListItems("dfe_qualification", "dfe_category"));
+            UpsertAndRemoveOrphanedTypes(crm.GetPickListItems("dfe_qualification", "dfe_type"));
+            UpsertAndRemoveOrphanedTypes(crm.GetPickListItems("dfe_candidatepastteachingposition", "dfe_educationphase"));
+            UpsertAndRemoveOrphanedTypes(crm.GetPickListItems("msevtmgt_event", "dfe_event_type"));
+            UpsertAndRemoveOrphanedTypes(crm.GetPickListItems("phonecall", "dfe_channelcreation"));
         }
 
-        private void UpsertModels<T>(IEnumerable<T> models, IQueryable<T> dbSet) where T : BaseModel
+        private void UpsertAndRemoveOrphanedModels<T>(IEnumerable<T> models, IQueryable<T> dbSet) where T : BaseModel
```
jebw, line 149: Really minor comment on naming - is UpsertAndRemoveOrphaned a long way of saying Synchronize? That said it is more explicit though?

ethax-ross: Ah! thank you - I couldn't see the woods for the trees on that one 😁

PR #81, opened 2020-06-09
Run location sync periodically

ethax-ross:
We currently seed the locations table with postcode geolocation data from a static CSV file we vendor in the source code. Postcodes will change over time, so this commit moves the location seeding to a background job that fetches the latest CSV, unzips it and adds any new postcodes into our database on a weekly basis.


GetIntoTeachingApi/Jobs/LocationSyncJob.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                File.Delete(zipPath);
+            }
+            
+            return Path.Combine(csvPath, UkPostcodeCsvFilename);
+        }
+
+        private static void DeleteCsv(string csvPath)
+        {
+            if (!Env.IsDevelopment) File.Delete(csvPath);
```
DrLazer, line 120: lets give that guy some brackets { }

GetIntoTeachingApi/Startup.cs
```diff
@@ -162,16 +162,25 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 
             app.UseAuthorization();
 
+            // Configure recurring jobs.
+            RecurringJob.AddOrUpdate<CrmSyncJob>("crm-sync", (x) => x.Run(), Cron.Daily());
+            RecurringJob.AddOrUpdate<LocationSyncJob>("location-sync", (x) => 
+                x.RunAsync("https://www.freemaptools.com/download/full-postcodes/ukpostcodes.zip"), Cron.Weekly());
+
             using (var serviceScope = app.ApplicationServices.CreateScope())
             {
```
DrLazer, line 171: does this particular 'using' need to be scoped?

ethax-ross: Nope; ReSharper picks up on them but not that one for some reason 🤷‍♂️

PR #87, opened 2020-06-11
Add support for Task-based asynchronous pattern

ethax-ross:
- Make store interactions utilise async/await for better throughput.
- Register CdsClientService as transient/cloned

The CdsServiceClient is not asynchronous and there is overhead in creating a new instance (as it will go off and authenticate then download the Dynamics metadata each time its created). To ensure that we can handle multiple async requests hitting the CRM this PR registers a singleton CdsServiceClient that we then clone as a transient dependency.

### DIFF
--- a/GetIntoTeachingApi/Services/Store.cs
+++ b/GetIntoTeachingApi/Services/Store.cs
@@ -25,11 +25,16 @@ namespace GetIntoTeachingApi.Services
             _dbContext = dbContext;
         }
 
-        public void Sync(ICrmService crm)
+        public async Task SyncAsync(ICrmService crm)
         {
-            SyncTeachingEvents(crm);
-            SyncPrivacyPolicies(crm);
-            SyncTypeEntities(crm);
+            var tasks = new []
+            {
+                SyncTeachingEvents(crm),
+                SyncPrivacyPolicies(crm), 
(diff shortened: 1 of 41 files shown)


GetIntoTeachingApi/Services/Store.cs
```diff
@@ -72,29 +75,15 @@ public IEnumerable<TeachingEvent> SearchTeachingEvents(TeachingEventSearchReques
...
-            {
-                var origin = CoordinateForPostcode(request.Postcode);
-
-                teachingEvents = teachingEvents.Where(te => te.Building != null && te.Building.Coordinate != null);
-
-                // Approximate distance filtering in the database, with a suitable error margin (treats distance as an arc degree).
-                teachingEvents = teachingEvents.Where(te => EarthCircumferenceInKm * 
-                    te.Building.Coordinate.Distance(origin) / 360 < request.RadiusInKm + ErrorMarginInKm);
+            if (request.Radius == null)
+                return await teachingEvents.ToListAsync();
```
DrLazer, line 84: brackets { }

ethax-ross: https://dfedigital.atlassian.net/browse/GITPB-328

PR #106, opened 2020-06-22
Update event registration to use manual matchback

ethax-ross:
Instead of matching the candidate automatically on event registration we want to manually make the user verify their email address first. The event registration endpoint now accepts a `candidateId` that can be used to add an existing candidate to an event (the id should be obtained from a matchback request and only used once a user has verified the email).

Also adds additional fields to the candidate entity to match the updated event registration process.

### DIFF
--- a/GetIntoTeachingApiTests/Models/TeachingEventRegistrationRequestTests.cs
+++ b/GetIntoTeachingApiTests/Models/TeachingEventRegistrationRequestTests.cs
@@ -1,0 +1,6 @@
+﻿namespace GetIntoTeachingApiTests.Models
+{
+    public class TeachingEventRegistrationRequestTests
+    {
+    }
+}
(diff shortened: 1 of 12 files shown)


GetIntoTeachingApiTests/Models/TeachingEventRegistrationRequestTests.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace GetIntoTeachingApiTests.Models
+{
+    public class TeachingEventRegistrationRequestTests
+    {
```
DrLazer, line 4: is this needed?

ethax-ross: It's not needed at the moment, but I've been creating the corresponding test files for all classes just so they are there, even if there is nothing to test yet (I'm not testing getters/setters).

PR #108, opened 2020-06-22
Event registration as a transaction

ethax-ross:
WIP PR to deep-insert the candidate with an event registration. This works fine for qualifications/past teaching positions, but for some reason on events it gives an error:

```
FaultException`1: The required contact reference is missing. Please provide a valid contact reference
```

It may be something to do with events being a built-in Dynamics type. I'm parking this for now as it's more of a nice-to-have and covers an edge-case condition (where the candidate insert succeeds but the event registration fails, retrying the job will create a duplicate candidate).

We can pick it back up again after the live beta.


ethax-ross:
Due to limitations/weirdness in the CRM this can't be done.

PR #109, opened 2020-06-23
Ignore inactive contacts and prioritise master records

ethax-ross:
- Rename GetCandidate to MatchCandidate

There is another `GetCandidate` method on the `ICrmService` that accepts a candidate ID; renaming this method to avoid confusion -- this is the method that should be used to 'matchback' an existing candidate.

- Ignore candidates with inactive statecode

According to the CRM team any 'merged' contact records will be set as inactive, so we don't need to explicitly check for the master record.

- Prioritise master records over merged children

Master records have a `merged` attribute of `false`; these should be prioritised over child records (that have been merged into a master record).


GetIntoTeachingApi/Services/CrmService.cs
```diff
@@ -16,6 +16,12 @@ public enum PrivacyPolicyType
             Web = 222750001,
         }
 
+        public enum CandidateStatus
+        {
+            Active = 0,
+            Inactive = 1,
```
DrLazer, line 22: you don't need the '= 1' it increments by each line as default, I know you love saving code ;)

PR #112, opened 2020-06-24
Convert API payloads to snake_case

ethax-ross:
The expectation is that Ruby applications will be consuming the API predominantly, so it makes sense to accept snake_case attributes instead of the default camelCase to make the implementation simpler on the service consuming the API.


ethax-ross:
We had a discussion about the merits of this and decided to leave it as camelCase and convert to snake_case in the client instead.

PR #118, opened 2020-06-24
Include tests in SonarCloud reporting

ethax-ross:


ethax-ross:
I'll take a look into that in a follow up PR 👍

PR #120, opened 2020-06-25
Add basic integration test

ethax-ross:
We will be end-to-end testing separately to ensure the integration with Dynamics is functioning correctly, however we won't run these tests as often as the unit/integration tests. This PR adds a basic integration test that will test the full request/response stack.

### DIFF
--- a/GetIntoTeachingApi/Models/MappingInfo.cs
+++ b/GetIntoTeachingApi/Models/MappingInfo.cs
@@ -13,6 +13,10 @@ namespace GetIntoTeachingApi.Models
         public IDictionary<string, IDictionary<string, string>> Relationships { get; set; } =
             new Dictionary<string, IDictionary<string, string>>();
 
+        public MappingInfo()
+        {
+        }
+
         public MappingInfo(Type type)
         {
             _type = type;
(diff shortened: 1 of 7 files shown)


GetIntoTeachingApi/Models/MappingInfo.cs
```diff
@@ -13,6 +13,10 @@ public class MappingInfo
         public IDictionary<string, IDictionary<string, string>> Relationships { get; set; } =
             new Dictionary<string, IDictionary<string, string>>();
 
+        public MappingInfo()
```
DrLazer, line 16: Is this needed?

ethax-ross: Yeah you need a parameterless constructor to be able to deserialise into the object

GetIntoTeachingApi/Utils/Env.cs
```diff
@@ -7,6 +7,7 @@ public class Env : IEnv
         public bool IsDevelopment => EnvironmentName == "Development";
         public bool IsProduction => EnvironmentName == "Production";
         public bool IsStaging => EnvironmentName == "Staging";
+        public bool IsTest => EnvironmentName == null;
```
DrLazer, line 10: is that a string check for null?  should it be string.IsNullOrEmpty() perhaps?

ethax-ross: It'll never be an empty string in this case; there's no really nice way of checking if we're running in the test environment vs dev, but we explicitly set the env name for all other environments so the `null` check works here

GetIntoTeachingApi/Models/MappingInfo.cs
DrLazer: ok cool

GetIntoTeachingApi/Utils/Env.cs
DrLazer: cool

PR #138, opened 2020-07-01
Scan API with OWASP

sfawcett123:
### Purpose
Trial of using the ZAP OWASP Scanner against the API

### Findings

1.  The scanner does not have an action out of the box for github
2. The scanner when configured seems to take too long to run, and maybe actually trying to execute the API, which could be problematic

### Conclusion
Do not run the ZAP OWASP Scanner on the API


sfawcett123:
Scanner does not work on API,

PR #162, opened 2020-07-15
Change RadiusInKm to method

ethax-ross:
Previously it was appearing in the Swagger documentation and apparently `JsonIgnore` doesn't work when `[FromQuery]` so instead of writing a custom schema filter to exclude it we can more easily exclude it by making it a method instead of a property.

### DIFF
--- a/GetIntoTeachingApi/Services/Store.cs
+++ b/GetIntoTeachingApi/Services/Store.cs
@@ -130,13 +130,13 @@ namespace GetIntoTeachingApi.Services
 
             // Approximate distance filtering in the database, with a suitable error margin (treats distance as an arc degree).
             teachingEvents = teachingEvents.Where(te => EarthCircumferenceInKm *
-                te.Building.Coordinate.Distance(origin) / 360 < request.RadiusInKm + ErrorMarginInKm);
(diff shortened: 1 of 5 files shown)


GetIntoTeachingApi/Services/Store.cs
```diff
@@ -130,13 +130,13 @@ public bool IsValidPostcode(string postcode)
...
             teachingEvents = teachingEvents.Where(te => EarthCircumferenceInKm *
-                te.Building.Coordinate.Distance(origin) / 360 < request.RadiusInKm + ErrorMarginInKm);
+                te.Building.Coordinate.Distance(origin) / 360 < request.RadiusInKm() + ErrorMarginInKm);
 
             var inMemoryTeachingEvents = await teachingEvents.ToListAsync();
 
             // Project coordinates onto UK coordinate system for final, accurate distance filtering.
             return inMemoryTeachingEvents.Where(te => te.Building.Coordinate.ProjectTo(DbConfiguration.UkSrid)
-                    .IsWithinDistance(origin.ProjectTo(DbConfiguration.UkSrid), (double)request.RadiusInKm * 1000));
+                    .IsWithinDistance(origin.ProjectTo(DbConfiguration.UkSrid), (double)request.RadiusInKm() * 1000));
```
DrLazer, line 139: hehe

PR #180, opened 2020-07-29
Added Simple Smoke test

sfawcett123:
Simple test script which will curl the application health check URL and check the SHA is what is expected.
Additionally it will check the status of the health check
All done in glorious technicolour

### DIFF
--- a/tests/confidence/healthcheck.sh
+++ b/tests/confidence/healthcheck.sh
@@ -15,8 +15,8 @@
 URL=${1}
 EXPECTED_SHA=${2}
 
-# URL="get-into-teaching-api-dev"
-# EXPECTED_SHA="730ff0a"
+#URL="get-into-teaching-api-dev"
+#EXPECTED_SHA="730ff0a"
 
 RED='\033[0;31m'
 GREEN='\033[0;32m'
@@ -24,8 +24,8 @@ NC='\033[0m'
 
 rval=0
 FULL_URL="https://${URL}.london.cloudapps.digital/api/operations/health_check"
-http_status=$(curl -s -I ${FULL_URL} | head -1 | cut -d$' ' -f2)
-if [ "${http_status}" != "405" ] 
+http_status=$(curl -o /dev/null -s -w "%{http_code}"  ${FULL_URL}) 
+if [ "${http_status}" != "200" ] 
 then
 	echo "${RED}HTTP Status ${http_status}${NC}"
 	rval=1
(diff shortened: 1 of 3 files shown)


tests/confidence/healthcheck.sh
```diff
@@ -0,0 +1,55 @@
...
+if [ "${http_status}" != "200" ] 
+then
+	echo "${RED}HTTP Status ${http_status}${NC}"
+	rval=1
+else
+	echo "${GREEN}HTTP Status is Healthy${NC}"
+
+        json=$(curl -s -X GET ${FULL_URL})
+        status=$( echo ${json} | jq -r .status)
+        if [ "${status}" != "healthy" ]
```
ethax-ross, line 37: This can also be 'degraded' as well as 'unhealthy', not sure if you want to differentiate here or not though

tests/confidence/healthcheck.sh
```diff
@@ -0,0 +1,55 @@
...
+FULL_URL="https://${URL}.london.cloudapps.digital/api/operations/health_check"
+http_status=$(curl -o /dev/null -s -w "%{http_code}"  ${FULL_URL}) 
+if [ "${http_status}" != "200" ] 
+then
+	echo "${RED}HTTP Status ${http_status}${NC}"
+	rval=1
+else
+	echo "${GREEN}HTTP Status is Healthy${NC}"
+
+        json=$(curl -s -X GET ${FULL_URL})
```
ethax-ross, line 35: If we did the verbose output on the first call we could work all this out from a single response, but I don't feel strongly about it

tests/confidence/healthcheck.sh
sfawcett123: I really can only handle a true / false state so if it is not healthy, then fail it, we can investigate. It is only being run after deployment and I would hope it was ok then

tests/confidence/healthcheck.sh
sfawcett123: Yes, agreed. but it's not a biggy :) the script can be improved as we move on. Thanks

tests/confidence/healthcheck.sh
ethax-ross: It may be worth checking for `unhealthy` explicitly in that case - we probably don't want to block the pipeline if it's degraded as that indicates a downstream service is having issues

sfawcett123: OK since I need to remove colour I will

PR #185, opened 2020-08-03
Increase TOTP window to 15 minutes

ethax-ross:
Currently the TOTP window is 30 seconds (+ the time remaining in the current 'step', which is a maximum of 30 seconds). It has been decided that this is too short and a window of 15 minutes is more suitable. This commit increases the window to 15 minutes (30 * 30 second 'steps').

The future window has been reduced from 1 to 0 as the server is both generating and consuming the TOTPs it should never be the case that a code is valid in a following step.


ethax-ross:
We can't do anything about the CVE failure at the moment: https://github.com/dotnet/aspnetcore/issues/23652

PR #196, opened 2020-08-07
Add PEN_TEST_SHARED_SECRET

ethax-ross:
In order to allow the penetration test team to connect to the API without exposing our `SHARED_SECRET` we are adding another valid secret that they can use.

This PR should be reverted once the penetration test is complete.

### DIFF
--- a/GetIntoTeachingApi/Auth/SharedSecretHandler.cs
+++ b/GetIntoTeachingApi/Auth/SharedSecretHandler.cs
@@ -1,4 +1,5 @@
-﻿using System.Security.Claims;
+﻿using System.Linq;
+using System.Security.Claims;
 using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using GetIntoTeachingApi.Utils;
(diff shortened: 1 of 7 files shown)


ethax-ross:
> Just a general observation here.
> Do we need to have a Utils/Env?
> Could this not all be moved into StartUp.cs and be handled with a .json settings file overidden at startup with environment vars?

What would be the benefit? It sounds more complex than what we have at the moment to me?

DrLazer:
> > Just a general observation here.
> > Do we need to have a Utils/Env?
> > Could this not all be moved into StartUp.cs and be handled with a .json settings file overidden at startup with environment vars?
> 
> What would be the benefit? It sounds more complex than what we have at the moment to me?

It's convention, and less complex because it automagically brings in your settings.  Less code

ethax-ross:
Created a ticket for picking this up in another PR: https://dfedigital.atlassian.net/browse/GITPB-572

PR #197, opened 2020-08-10
Disable ETag cache in development

ethax-ross:
Having the ETag caching enabled in development can make it harder to debug issues; particularly around the store sync process - if the sync fails and you hit the endpoint expecting to get types and it then returns an empty response, that will potentially remain cached even if you fix the bug.

### DIFF
--- a/GetIntoTeachingApi/Attributes/CrmETagAttribute.cs
+++ b/GetIntoTeachingApi/Attributes/CrmETagAttribute.cs
@@ -4,6 +4,7 @@ using System.Security.Cryptography;
 using System.Text;
 using GetIntoTeachingApi.Jobs;
 using GetIntoTeachingApi.Services;
+using GetIntoTeachingApi.Utils;
 using Hangfire;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Filters;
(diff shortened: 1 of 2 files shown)


GetIntoTeachingApi/Attributes/CrmETagAttribute.cs
```diff
@@ -14,22 +15,26 @@ public class CrmETagAttribute : Attribute, IActionFilter
...
             _hangfireJobStorage = hangfireJobStorage;
             _metrics = metrics;
+            _env = env;
         }
 
         public void OnActionExecuting(ActionExecutingContext context)
         {
-            var canCacheRequest = context.HttpContext.Request.Method == "GET";
+            var isGetRequest = context.HttpContext.Request.Method == "GET";
+            var canCacheRequest = isGetRequest && !_env.IsDevelopment;
```
jebw, line 37: Does this need to handle test environments as well? Presumably caching should be disabled in those?

ethax-ross: Yeah that's a good shout; at the moment I mock it to disable caching but that would be better as the default behaviour - I'll update this in a bit

PR #210, opened 2020-08-13
Add WebFeedId to TeachingEvent and prevent sign ups if null

ethax-ross:
- Add WebFeedId to TeachingEvent

- Only accept online registrations for applicable events

If an event does not have a `WebFeedId` it is not possible to register for this event via the GITIS CRM. This commit adds a relevant validation check to reject such payloads from the client apps.

### DIFF
--- a/GetIntoTeachingApi/Models/TeachingEvent.cs
+++ b/GetIntoTeachingApi/Models/TeachingEvent.cs
@@ -23,6 +23,10 @@ namespace GetIntoTeachingApi.Models
         public int StatusId { get; set; }
         [EntityField("msevtmgt_readableeventid")]
         public string ReadableId { get; set; }
+        [EntityField("dfe_eventwebfeedid")]
+        [SwaggerSchema("If set, the API will accept new attendees for " +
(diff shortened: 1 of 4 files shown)


sfawcett123:
[e.g. Book flights](https://trello.com/c/gFLtDH7X/1-eg-book-flights)

GetIntoTeachingApi/Models/TeachingEvent.cs
```diff
@@ -23,6 +23,10 @@ public enum Status
         public int StatusId { get; set; }
         [EntityField("msevtmgt_readableeventid")]
         public string ReadableId { get; set; }
+        [EntityField("dfe_eventwebfeedid")]
+        [SwaggerSchema("If set, the API will accept new attendees for " +
+            "this event (an external sign up should be used if this value is nil).")]
```
DrLazer, line 28: Why is the string here concatenated?

ethax-ross: So it fits on my screen 🙄

PR #217, opened 2020-08-14
Move de-duplication logic into TeachingEventRegistration model

ethax-ross:
Previously we saved the `TeachingEventRegistration` as part of saving a `Candidate`, but this was later changed so that the registration is saved independently (a deep insert on `Candidate` doesn't work for reasons unknown).

The de-duplication logic was left in the `Candidate` model - this moves it into the `TeachingEventRegistration` model, so if a registration already exists for the user against the event a call to `ToEntity` returns `null` and the `Save` operation on the `CrmService` just returns.

### DIFF
--- a/GetIntoTeachingApi/Models/TeachingEventRegistration.cs
+++ b/GetIntoTeachingApi/Models/TeachingEventRegistration.cs
@@ -33,5 +33,17 @@ namespace GetIntoTeachingApi.Models
             : base(entity, crm)
         {
         }
+
+        protected override bool ShouldMap(ICrmService crm)
+        {
+            var alreadyRegistered = !crm.CandidateYetToRegisterForTeachingEvent(CandidateId, EventId);
+
+            if (alreadyRegistered)
+            {
+                return false;
+            }
+
+            return base.ShouldMap(crm);
+        }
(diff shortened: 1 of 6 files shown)


GetIntoTeachingApi/Models/TeachingEventRegistration.cs
```diff
@@ -33,5 +33,17 @@ public TeachingEventRegistration(Entity entity, ICrmService crm)
             : base(entity, crm)
         {
         }
+
+        protected override bool ShouldMap(ICrmService crm)
+        {
+            var alreadyRegistered = !crm.CandidateYetToRegisterForTeachingEvent(CandidateId, EventId);
```
DrLazer, line 39: dont need the variable here do we?

if(!thething())
{
    return false;
}

ethax-ross: It was just to make it more readable as it's negating the function call - I could have reworked the function to something like `crm.CandidateAlreadyRegisteredForTeachingEvent` but there are multiple in this format so opted to keep them consistent (and the current method name makes more sense in the other use cases).

PR #231, opened 2020-08-20
Temporarily disable sending data to the CRM

ethax-ross:
We are conducting a UX session on Wednesday and don't want to collect any personally identifiable information that then gets sent to the CRM. The easiest way to prevent this is to disable the API from writing any candidate information to the CRM.


sfawcett123 (approved):
LGTM

DrLazer (approved):

PR #235, opened 2020-08-27
Convert all DateTime instances to be UTC

ethax-ross:
The API should be time zone agnostic; currently most of the date instantiations use `DateTime.Now`, which returns a date in the time zone of the system. Instead, this commit changes to use `DateTime.UtcNow`, which will return a UTC date that is consistent with what the CRM expects/uses.


DrLazer (dismissed):
looks good

DrLazer (approved):

PR #236, opened 2020-08-28
Return OnlineEvent type events in distance based searches

ethax-ross:
When a user searches for events that are a certain distance from a postcode we currently do not return events that have the `OnlineEvent` type. This PR updates the search query to return online events events for distance based searches.

It performs a second query to retrieve them (concatenating the results); whilst it would be possible to do it in a single query it would be more verbose (we would need to use LINQKit `PredicateBuilder`), and given the number of events we are expecting there shouldn't be a noticeable performance hit for doing two queries.


DrLazer (approved):
looks good

PR #237, opened 2020-08-28
Changes to the event sign up when opting-in to the mailing list

ethax-ross:
- Make PreferredTeachingSubjectId required

In the client this is no longer an optional field, so this will enforce the validation in the API.

- Add additional fields to event sign up

When signing up for an event, if a candidate opts-in to the mailing list we will be asking them the 3 additional questions from the mailing list sign up:

* What stage are you at with your degree?
* How close are you to applying for teacher training?
* What subject do you want to teach?

This commit adds the corresponding fields to the `TeachingEventAddAttendee` model and makes them required if the candidate has opted-in to the mailing list.

- Changes to opt-in to event information on mailing list sign up

We no longer have an 'opt in to event information' question on the client; instead we should opt-in the user to event information if they opt-in to the mailing list and provide an address postcode.

- Use correct events subscription type

We now want to always create an event subscription when signing up to the mailing list.

The postcode is optional on the mailing list and events sign up, however if they provide a postcode then the event subscription type should be 'local', otherwise it defaults to 'single'.

---

The models for signing up for the mailing list and events are getting very similar now; it may be worth combining these in the future into a single model used for both actions.

### DIFF
--- a/GetIntoTeachingApi/Models/Candidate.cs
+++ b/GetIntoTeachingApi/Models/Candidate.cs
@@ -63,6 +63,7 @@ namespace GetIntoTeachingApi.Models
         public enum SubscriptionType
         {
             SingleEvent = 222750001,
+            LocalEvent = 222750000,
         }
 
         public enum PreferredEducationPhase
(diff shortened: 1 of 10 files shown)


jebw (approved):
LGTM

PR #238, opened 2020-09-01
Remove unused JsonIgnore attributes

ethax-ross:
The Candidate/PhoneCall/CandidatePrivacyPolicy models are no longer directly exposed to the client (instead we return explicit response models) - we no longer need the `JsonIgnore` attributes to prevent fields from being serialized.


jebw (approved):
LGTM

PR #240, opened 2020-09-01
Changed status Cake check frequency to every minute

sfawcett123:
Increased the frequency from 300 ( 5 mins ) to 60 ( 1 Min )
Removed need for HTTP_USERNAME 
Added Github workflow to install


jebw (approved):
LGTM

I can't see any runs for this in the Actions tab? Then again I can't see how to run it without testing first either

sfawcett123:
> LGTM
> 
> I can't see any runs for this in the Actions tab? Then again I can't see how to run it without testing first either

This is a major problem with github actions, they only seem to appear when on the Master Branch.  The way I have been working is testing it on a copy of the repository or using a tool called ACT. ACT though isn't great

ethax-ross (approved):

PR #242, opened 2020-09-02
Only expose SwaggerUI on non-production environments

ethax-ross:
The SwaggerUI exposes the internal IP address of the host machine, so we only want this to be accessible on non-production environments to facilitate testing.


jebw (approved):
LGTM

PR #251, opened 2020-09-17
Fix exception on retrieving callback booking quotas

ethax-ross:
When querying callback booking quotas attempting to do the availability check as part of the Dynamics query throws an exception; I'm not sure why, but moving this to be performed in-memory resolves the issue.


DrLazer (approved):
looks good

PR #253, opened 2020-09-18
Commit encrypted credentials to repository

ethax-ross:
As there's not an easy way of managing per-environment credentials in CI we're going to commit the encrypted credentials to the repository using `git-secret`. It's then just a case of revealing the appropriate credential for the environment, so:

```
cd GetIntoTeachingApi
SECRETS_DIR=.gitsecret-production tell new@user.com
SECRETS_DIR=.gitsecret-production git-secret reveal
```


ethax-ross:
After talking it through with other members of dev ops we're opting to store the credentials in Azure Vault or AWS and pipe them in via terraform.

PR #254, opened 2020-09-20
Initial Production Build

sfawcett123:
The API is required for all of the services to run, and therefore it's deployment has been designated master. 
Thus this is where we deploy logging, and other common resources from.
Once the API has been sucessfully tested the TTA and APP can layer on top


sfawcett123:
I know the state file is being held in the test environment, this is because we have not been granted authority to use our own production Azure environment just yet. Although we have got access to deploy to our container environment. 
Since the state is secure where it is and can be moved, we feel we can progress

PR #267, opened 2020-09-29
Log request/response with payloads

ethax-ross:
- Add filter to log request/response paylaods

We want to be able to see request/response payloads to help us when debugging requests in production.

Add `LogRequests` attribute that will log request/response information per-controller and enable for every controller.

Add `LoggableAttribute` to opt-in models that should be logged.

Add `SensitiveDataAttribute` to opt-out any sensitive attributes from being logged.

- Make request/response models loggable

Add the `Loggable` attribute to request/response models.

Add the `SensitiveData` attribute to any attributes within the models that we want to omit from the logs.

### DIFF
--- a/GetIntoTeachingApi/Models/TeacherTrainingAdviserSignUp.cs
+++ b/GetIntoTeachingApi/Models/TeacherTrainingAdviserSignUp.cs
@@ -1,10 +1,12 @@
 ﻿using System;
 using System.Linq;
 using System.Text.Json.Serialization;
+using GetIntoTeachingApi.Attributes;
 using Swashbuckle.AspNetCore.Annotations;
 
 namespace GetIntoTeachingApi.Models
 {
+    [Loggable]
     public class TeacherTrainingAdviserSignUp
     {
         public Guid? CandidateId { get; set; }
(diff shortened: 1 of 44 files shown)


GetIntoTeachingApi/Models/TeacherTrainingAdviserSignUp.cs
```diff
@@ -25,17 +27,26 @@ public class TeacherTrainingAdviserSignUp
...
+        [SensitiveData]
         public string TeacherId { get; set; }
         public string DegreeSubject { get; set; }
+        [SensitiveData]
         public string Telephone { get; set; }
+        [SensitiveData]
         public string AddressLine1 { get; set; }
+        [SensitiveData]
         public string AddressLine2 { get; set; }
         public string AddressCity { get; set; }
```
jebw, line 48: Any reason not to include AddressCity as well?

jebw (approved):
Looks good, I did notice the AddressCity wasn't decorated - not sure whether that's by design or because it got missed?

GetIntoTeachingApi/Models/TeacherTrainingAdviserSignUp.cs
ethax-ross: I figured that `AddressCity` isn't personally identifiable (arguably `AddressPostcode` as well but I think some postcode regions can be very small) and it may help with debugging if left in -- happy to take it out if you disagree, but I'll do that in a follow up PR 👍

PR #270, opened 2020-09-30
Update API Grafana Dashboard

ethax-ross:
Group metrics by rows.

Add engagement and health metrics.

Select Elasticsearch source via variable (we're looking at ways to inject this from the environment as part of CI).


sfawcett123 (approved):
LGTM

PR #283, opened 2020-10-05
Use recommended Sentry integration method

ethax-ross:
The Sentry C# client appears to recommend a different way of integrating it to what Sentry's wizard shows you initially; this updates to leverage `useSentry()` instead of `Sentry.Init` directly.

Also includes temporary operations endpoint to enable testing the integration in production.


sfawcett123 (approved):
Have you updated the SENTRY_DSN Secret?

ethax-ross:
Yep, I'll remove the legacy `SENTRY_URL` once this has shipped.

PR #285, opened 2020-10-05
Product Release Workflow to Permit TAG

sfawcett123:
When you release to production it will ask for a TAG, this tag should be the Tag for the Release you want to send to Production.
This will then send the version of the Docker Image and Terraform to Production rather than the current master.

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -1,7 +1,13 @@
 ---  
 name: Release to Production
 on: 
+   repository_dispatch:
    workflow_dispatch:
+    inputs:
+      tags:
+        description: 'Release Tag'
+        required: true
+        default: 'master'
 
 env:
   DOCKERHUB_REPOSITORY:     dfedigital/get-into-teaching-api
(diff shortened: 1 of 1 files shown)


.github/workflows/production.yml
```diff
@@ -1,7 +1,13 @@
 ---  
 name: Release to Production
 on: 
+   repository_dispatch:
    workflow_dispatch:
+    inputs:
+      tags:
+        description: 'Release Tag'
+        required: true
+        default: 'master'
```
jebw, line 10: Can this be blank, or the default not set. 

I'd like to see releasing from `master` _very_ strongly discouraged - we should always be releasing a tag from testing

jebw (changes requested):

.github/workflows/production.yml
sfawcett123: I am going to remove it as you suggest and I am going to also add a check so it errors if you do chose master. I will look at a way to see if the tag is in the list of releases, so you cannot choose any old branch

jebw (approved):
LGTM

PR #287, opened 2020-10-06
Statuscake Terraform Tidyup of format

sfawcett123:
Ran 'terraform fmt' on terraform files to improve their format


jebw (changes requested):
This appears to be whitespace changes rather than changing the timeout?

jebw (approved):

PR #288, opened 2020-10-06
Workaround for exception in Prometheus.HttpMetrics.HttpRequestDurationMiddleware

ethax-ross:
It appears to be fixed in a later version of .NET Core but the version we are using occasionally throws an `InvalidOperationException` due to performing a synchronous operation.

This updates Kestrel config to allow sync operations.


sfawcett123 (approved):
Good

PR #291, opened 2020-10-07
Set CdsServiceClient timeout to 30s

ethax-ross:
The `CdsServiceClient` timeout by default is 2 minutes, but the healthcheck endpoint depends on the CRM and we want it to complete in less time than that if there is a problem.


jebw (approved):
LGTM

PR #295, opened 2020-10-10
Supress StyleCop warning for migration

ethax-ross:
The migrations are auto-generated, so we don't want to be warned of StyleCop violations.


peteryates (approved):
👍

PR #296, opened 2020-10-10
Changes to Subscription modelling

ethax-ross:
- Always opt in to bulk emails/mm on event sign up

When a candidate signs up for an event we want to be able to both send them bulk emails and marketing material.

- Prevent event subscription being downgraded

When a candidate signs up for an event and passes their post code when opting into the mailing list, we give them a 'local event' subscription type. If they then go on to sign up for another event and don't provide their postcode we currently downgrade their subscription to a 'single event' type.

Instead, if they have provided their postcode as part of an event sign up in the past then we want to keep their subscription type as 'local event'.


peteryates (approved):
I've read through and everything looks sensible. I need to improve my familiarity with the codebase and API though.

PR #297, opened 2020-10-10
Opt-out of DoNotBulkEmail on event sign up

ethax-ross:
When signing up for an event subscription we should be setting the contact-level `DoNotBulkEmail` preference to `false`.


GetIntoTeachingApi/Models/TeachingEventAddAttendee.cs
```diff
@@ -117,7 +117,7 @@ private void AddQualification(Candidate candidate)
         private void ConfigureConsent(Candidate candidate)
         {
             candidate.OptOutOfSms = false;
-            candidate.DoNotBulkEmail = !SubscribeToEvents && !SubscribeToMailingList;
+            candidate.DoNotBulkEmail = false;
```
ethax-ross, line 120: Ensuring this is `false` is covered in another test.

peteryates (commented):
Looks good.

PR #298, opened 2020-10-13
Upgrade Monitor Terraform to 13.4

sfawcett123:
This only upgrades the Terraform for the Monitoring Deployment and does not effect delivery of the application.


terraform/monitoring/modules.tf
```diff
@@ -0,0 +1,60 @@
...
+  monitoring_space_id      = data.cloudfoundry_space.space.id
+  monitoring_org_name      = local.monitoring_org_name
+  environment_variable_map = local.template_variable_map
+}
+
+
+module "grafana" {
+  source                  = "git::https://github.com/DFE-Digital/bat-platform-building-blocks.git//terraform/modules/grafana?ref=monitoring-terraform-0_13"
+  monitoring_space_id     = data.cloudfoundry_space.space.id
+  monitoring_org_name     = "${var.environment}-${var.grafana["name"]}"
```
vigneshmsft, line 51: does `prometheus["name"]` and `grafana["name"]` have to different?
can we just use `local.monitoring_org_name` here and remove `name` key from promethus and grafana var maps if they aren't used?

terraform/monitoring/provider.tf
```diff
@@ -4,4 +4,15 @@ provider "cloudfoundry" {
...
+  required_version = ">= 0.13.4"
+
+  required_providers {
+    cloudfoundry = {
+      source  = "cloudfoundry-community/cloudfoundry"
+      version = "0.12.6"
+    }
+  }
+}
+
```
vigneshmsft, line 17: extra new line? also line 58 in modules.tf file

terraform/monitoring/variables.tf
```diff
@@ -66,7 +66,7 @@ variable "alertmanager" {
   }
 }
 
-variable "google_client_id" {}	
+variable "google_client_id" {}
```
vigneshmsft, line 69: remove '"' from variable names and keep it consistent (as we are touching this file)?

vigneshmsft (approved):

terraform/monitoring/modules.tf
sfawcett123: No I just have them different though it might be a good idea to rationalise one day

terraform/monitoring/provider.tf
sfawcett123: ok .. next time it is changed I will pickit up

terraform/monitoring/variables.tf
sfawcett123: Next time it is changed I will pickit up

PR #308, opened 2020-10-19
Trim white space from JSON string attributes

ethax-ross:
When a HTTP request comes in to ASP.NET Web API the payload is bound to a request model; when this happens we hook into the operation and ensure any string attributes are trimmed.

When we read/write JSON string attributes there aren't any instances where having trailing/leading white space would be useful or intentional.

Accepting white space on the start/end of strings is currently causing issues with the postcode validator, as the regex is strict in this respect, yet it shouldn't matter if a user enters a postcode with white space around it.

Initially I attempted doing this with a custom `JsonConverter`, but that only applied to POST request body payloads and not GET query parameters. By using a model binder we should be covered for any content-type.


peteryates (approved):
Makes sense :+1:

PR #315, opened 2020-10-22
Accept X-Client-IP header on all requests

ethax-ross:
Update the Swagger configuration to accept an `X-Client-IP` header on all requests. The auto-generated library will allow this to be specified per-API request, for example:

```
result = api_instance.get_callback_booking_quotas(x_client_ip: "1.2.3.4")
```

Going forward this IP address will be used to rate limit requests to the API, in addition to the API key of the client (we can create custom IP/Client ID resolvers in the [ASPNetCoreRateLimit](https://github.com/stefanprodan/AspNetCoreRateLimit) package).

If the IP address is unspecified the requests will fall into the rate limiting bucket based solely on the client ID (API key in use).

Can be set to the end user IP address to allow per-user rate limiting in the API. This is how we will be rate limiting individual users when they request a PIN code email (so that they cannot abuse the GOV.UK Notify service to send spam emails).


peteryates (approved):
Looks good.

PR #319, opened 2020-10-23
Cache responses from Google Geocoding API

ethax-ross:
When we can't find a postcode geolocation from our local cache we go off to the Google Geocoding API. We're finding that many teaching events share the same postcode, so we're requesting it multiple times from Google during a sync, which both slows the sync down and costs us money. Instead, this commit persists the location to our cache so that subsequent lookups don't  need to hit Google's API.


peteryates (approved):
Nice.

PR #323, opened 2020-10-27
Add endpoint to get upcoming events by type

ethax-ross:
- Add endpoint to get upcoming events by type

When a user lands on the events index page in the website we want to be able to show them the next 3 events that are upcoming for each type. This improves the user experience when we are approaching the end of the month and there are no
longer any events within the current month to display.

Adds a new endpoint `/api/teaching_events/upcoming_indexed_by_type` that will return the upcoming teaching events grouped by their type.

The return type is `Dictionary<string, IEnumerable<TeachingEvent>>` due to `System.Text.Json` not yet supporting non-string Dictionary keys (it would be preferable to use an `int` here).

Adds a `/api/teaching_events/search_indexed_by_type` for consistency, with a view to removing the current `/search` endpoint once deprecated in the front end.

```
{
  "222750001": [
    {
      "typeId": 222750001,
      "id": "1d3d055a-84dd-ea11-a813-000d3a44afcc",
      ...
    },
    {
      "typeId": 222750001,
      "id": "293d055a-84dd-ea11-a813-000d3a44afcc",
      ...
    }
  ],
  "222750008": [
    {
      "typeId": 222750008,
      "id": "f5a82568-f402-eb11-a812-000d3a44af3f",
      ...
    },
    ...
  ]
}
```

### DIFF
--- a/GetIntoTeachingApi/Controllers/TeachingEventsController.cs
+++ b/GetIntoTeachingApi/Controllers/TeachingEventsController.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using GetIntoTeachingApi.Attributes;
 using GetIntoTeachingApi.Jobs;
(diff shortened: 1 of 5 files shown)


peteryates (approved):
This looks great.

PR #326, opened 2020-10-28
Rate limit vulnerable endpoints by client 

ethax-ross:
- Enforce rate limiting to vulnerable endpoints

Rate limits endpoints that could be exploited:

```
POST /candidates/access_tokens
POST /mailing_list/members
POST /teaching_events/attendees
POST /teacher_training_adviser/candidates
```

Set to 30 requests/min which is well above the current traffic (peaking at ~2rpm). This should have enough buffer in it for when we go up to 40% rollout.

The /operations endpoints are excluded from rate limiting.

- Add graph to monitor rate limited requests

Adds a graph to display the requests/minute for each of the rate limited endpoints. The panel also has thresholds for 15rpm (warning) and 30rpm (critical) so that we can see when traffic is approaching the rate limits and needs to be adjusted.

Screenshot 2020-10-28 at 15 27 05

### DIFF
--- a/GetIntoTeachingApi/Startup.cs
+++ b/GetIntoTeachingApi/Startup.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Linq;
+using AspNetCoreRateLimit;
 using FluentValidation.AspNetCore;
 using GetIntoTeachingApi.Adapters;
 using GetIntoTeachingApi.Auth;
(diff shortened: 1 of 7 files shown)


GetIntoTeachingApi/Startup.cs
```diff
@@ -230,5 +236,30 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment hostEnv)
...
             });
         }
+
+        private void ConfigureRateLimiting(IServiceCollection services)
+        {
+            // Load appsettings.json.
+            services.AddOptions();
+
+            // Stores counters/IP rules.
+            services.AddMemoryCache();
```
jebw, line 246: Is there an option to use a Redis cache - in theory load balancing and not much other load you'll end up applying N * limits you've set - where N is the number of nodes

jebw: Or memcache?

jebw: Scratch that - just read the README, - RTFM, lol

jebw (approved):
LGTM

PR #327, opened 2020-10-29
Remove deprecated search endpoint

ethax-ross:
This endpoint was replaced with `search_indexed_by_type`.

Updates the test for the ETag attribute to include the two new search endpoints (I had missed this in the original PR).


peteryates (approved):
Makes sense.

PR #331, opened 2020-10-30
Add alert rules for any 429s and high 404s

ethax-ross:
As a first attempt at getting some metric alerting in place, this should inform us when a client or end user receives a 429 response (too many requests) or if there are high numbers of 404s being triggered in the app/tta services.


peteryates (approved):
Looks like it makes sense.

PR #341, opened 2020-11-02
Monitoring/alertmanager/slack

sfawcett123:
Add Slack to the alertmanager


ethax-ross (approved):
Looks good, I think we should have a dedicated alerts channel going forward but one to bring up with the rest of the team.

PR #346, opened 2020-11-02
Log Google API response status text

ethax-ross:
We are seeing an error on dev and the API key has been checked, so it must be something else going on. This should give us the status text back from Google.


peteryates (approved):
:+1:

PR #348, opened 2020-11-03
Make entrypoint.sh executable

ethax-ross:
Docker copies the file but can't run it as it doesn't have executable permissions; making it executable here should mean docker mimics the permissions on `COPY`.


sfawcett123 (approved):
The file is empty?

ethax-ross:
It's the permissions of the file that have changed

PR #351, opened 2020-11-03
Run migrations earlier

ethax-ross:
We are getting permission errors occasionally on deployment due to migrations running on multiple instances. To reduce of the chance of this happening this commit will run migrations earlier (hopefully before the application has fully 'started' so before the second instance gets spun up) and also check to ensure pending migrations exist before attempting to run them.

### DIFF
--- a/GetIntoTeachingApi/Database/DbConfiguration.cs
+++ b/GetIntoTeachingApi/Database/DbConfiguration.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


peteryates (approved):
On the face of it this looks good. Would it make sense to give the instances roles so one could be responsible for always running the migrations? That would remove any chance element.

Either way, this still looks like an improvement

ethax-ross:
Good suggestion, I'd need to look into how possible that is with GOV PaaS, we'd also need to ensure that the instance responsible for running the migration boots first. I'm hopeful this will be good enough, as long as the migrations are ran before the instance has 'started' (and therefore the next instance begins to spin up).

PR #352, opened 2020-11-03
Ignore leading/trailing whitespace in match-back

ethax-ross:
The CRM contains some candidates that have leading/trailing white space around their first name, last name and email address.

Ignore white space when comparing first/last name and email address.

Match on emails containing the provided email string initially (to positively match even if the email in the CRM has white space) - we do another, exact match as part of the in-memory match-back.


peteryates (approved):
Looks good to me.

PR #353, opened 2020-11-03
Disable rate limiting in Staging environments

ethax-ross:
The API uses the staging environment for all production-like environments; we don't want to enforce rate limiting in these environments so that the integration tests can successfully run.


peteryates (approved):
Makes sense.

PR #357, opened 2020-11-05
Only run migrations on first instance

ethax-ross:
We are getting conflicts because two instances are trying to run the migrations at the same time.

Restrict to instance 0 running the migrations, which should be the first instance spun up in a deployment.


sfawcett123 (approved):
LGTM

PR #358, opened 2020-11-09
Made changes to bring in line with Becoming A teacher

sfawcett123:
The central repository of Becoming A Teacher building blocks has changed with some of the features developed for GiT being made available to other teams.
Since these features are now in the central repository they can be removed from this local version


ethax-ross (approved):
Looks good, assuming the test alert is to stay in?

PR #361, opened 2020-11-10
Track how long UpsertCandidateJobs spend in the queue

ethax-ross:
Add a Histogram metric to track the job duration for `UpsertCandidateJob` so that we can get a feel for how long they are remaining in the queue for before being sent to the API or dropped.


jebw (approved):
LGTM

PR #362, opened 2020-11-10
Default CF_INSTANCE_INDEX to 0

ethax-ross:
If we are running locally we always want to run the migrations. In the event that the env var is not set we would also be better off running the migrations than not, so defaulting to 0 makes sense.


sfawcett123 (approved):
LGTM

PR #367, opened 2020-11-11
Link API alert rules to the runbook/Grafana panels

ethax-ross:
Link the API alert rules to the runbook in Confluence and the respective Grafana panels.


sfawcett123 (approved):
LGTM

PR #368, opened 2020-11-11
Link App/TTA alert rules to the runbook/Grafana

ethax-ross:
Link the App and TTA alert rules to the runbook in Confluence and the respective Grafana panels.


sfawcett123 (approved):
LGTM

PR #369, opened 2020-11-11
Increase rate limit thresholds

ethax-ross:
- Increase rate limit thresholds

We saw a spike of match back requests nearing 60rpm, so upping it to 250rpm and increasing the other endpoints to 100rpm.

- Update thresholds to match new rate limits

- Split rate limit alert rules

We now have two rules, one to cover the majority of sign up requests and one for the match back endpoint, which receives more requests.


jebw (approved):

peteryates (approved):
Looks sensible

PR #370, opened 2020-11-11
Fix Google API calls panel

ethax-ross:
This had been set to the random walk metric for some reason.

Also shuffles some of the panels around, which is why the diff is so big.


sfawcett123 (approved):
LGTM

PR #375, opened 2020-11-16
Switch to docker-compose for development

ethax-ross:
- Run against Postgres DB in development/test

Updates the development and test environments to execute against a dockerized Postgres database instance instead of using the in-memory database.

The advantages to do this are:

- Migrations are ran in the test suite/development.
- We can load and persist all the geolocation data in development.
- Development/test more closely match production.
- It's easier to run the app on a non-Windows machine.

The test environment works by creating a new database for each test that requires one; we use the `gis_test` database that is created for us as a template (after migrations have been ran).

Whilst it would be quicker to wrap each test run in a transaction using the same database, this setup would be more complicated and using Postgres templates seems to be fast enough for now.

- Remove in-memory SQLite references

Removes the in-memory SQLite dependencies.

The rate limiting integration tests hit the `Startup` class, so the `WebApplicationFactory` also has to be tweaked to cut out the default `DbContext` and replace it with an instance that points to the database that was created for the unit test being ran.

- Spin up test stack in CI

Run `docker-compose up` as part of CI so that the test database and any other resources are available.

- Tweaks to make development setup easier

Default some of the env variables that are not sensitive/don't need to change.

Clean up `DatabaseTests`.

### DIFF
--- a/GetIntoTeachingApiTests/Helpers/DatabaseTests.cs
+++ b/GetIntoTeachingApiTests/Helpers/DatabaseTests.cs
@@ -7,23 +7,15 @@ namespace GetIntoTeachingApiTests.Helpers
 {
     public abstract class DatabaseTests : IDisposable
     {
-        public GetIntoTeachingDbContext DbContext { get; }
+        public GetIntoTeachingDbContext DbContext { get; set; }
 
         protected DatabaseTests(DatabaseFixture databaseFixture)
         {
             var id = Guid.NewGuid().ToString().Replace("-", "");
             var databaseName = $"gis_test_{id}";
-            var connectionString = $"Host=localhost;Database={databaseName};Username=docker;Password=docker";
 
-            using var templateConnection = new NpgsqlConnection(databaseFixture.ConnectionString);
-            templateConnection.Open();
-
-            using var command = new NpgsqlCommand($"CREATE DATABASE {databaseName} WITH TEMPLATE {databaseFixture.TemplateDatabaseName};", templateConnection);
-            command.ExecuteNonQuery();
-                
-            var builder = new DbContextOptionsBuilder<GetIntoTeachingDbContext>();
-            DbConfiguration.ConfigPostgres(connectionString, builder);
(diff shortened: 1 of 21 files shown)


GetIntoTeachingApiTests/Helpers/DatabaseTests.cs
```diff
@@ -1,30 +1,52 @@
...
+
+            DbContext = new GetIntoTeachingDbContext(builder.Options);
+        }
+
+        private static void CloneTemplateDatabase(DatabaseFixture databaseFixture, string databaseName)
+        {
+            using var templateConnection = new NpgsqlConnection(databaseFixture.ConnectionString);
+            templateConnection.Open();
+
+            using var command = new NpgsqlCommand($"CREATE DATABASE {databaseName} WITH TEMPLATE {databaseFixture.TemplateDatabaseName};", templateConnection);
```
peteryates, line 48: I've never seen the `TEMPLATE` syntax with the word `WITH` in it, guessing it's optional. Interesting!

docker-compose.yml
```diff
@@ -0,0 +1,16 @@
+version: "3.8"
+services:
+  postgres:
+    image: kartoza/postgis:12.0
```
peteryates, line 4: Is there any reason to use this one over the official `postgis/postgis` image?

peteryates (approved):
Looks good to me, `docker-compose up` works well locally. Couple of super-minor comments/observations.

docker-compose.yml
ethax-ross: I mainly went with this one because it supports SSL out of the box, so I don't need to faff around setting it up manually or toggle the flag when we're not running in production. It also supports booting up multiple databases as part of the environment config which was nice

GetIntoTeachingApiTests/Helpers/DatabaseTests.cs
ethax-ross: Heh yeah, I've only ever used like this but it does look like it's optional

PR #380, opened 2020-11-24
Update event search validation to allow area only postcode

jkempster34:
- Update regex for postcodes
- Add to list of postcode parameters in validation tests
- Add store test for outward postcode

### DIFF
--- a/GetIntoTeachingApi/Models/Location.cs
+++ b/GetIntoTeachingApi/Models/Location.cs
@@ -9,8 +9,14 @@ namespace GetIntoTeachingApi.Models
 {
     public class Location
     {
+        public static readonly Regex OutwardOrFullPostcodeRegex = new Regex(
+            $"^({OutwardPostcodePattern}|{FullPostcodePattern})$", RegexOptions.IgnoreCase);
         public static readonly Regex PostcodeRegex = new Regex(
-            @"^([A-Z][A-HJ-Y]?\d[A-Z\d]? ?\d[A-Z]{2}|GIR ?0A{2})$", RegexOptions.IgnoreCase);
+            $"^({FullPostcodePattern})$", RegexOptions.IgnoreCase);
+
+        private const string OutwardPostcodePattern = @"[A-Z][A-HJ-Y]?\d[A-Z\d]?";
+        private const string InwardPostcodePattern = @"\d[A-Z]{2}|GIR ?0A{2}";
+        private const string FullPostcodePattern = OutwardPostcodePattern + " ?" + InwardPostcodePattern;
 
         [Key]
         public string Postcode { get; set; }
--- a/GetIntoTeachingApi/Models/Validators/TeachingEventSearchRequestValidator.cs
+++ b/GetIntoTeachingApi/Models/Validators/TeachingEventSearchRequestValidator.cs
@@ -16,7 +16,7 @@ namespace GetIntoTeachingApi.Models.Validators
             RuleFor(request => request.Postcode)
                 .NotEmpty()
                 .MaximumLength(40)
-                .Matches(Location.PostcodeRegex)
+                .Matches(Location.OutwardOrFullPostcodeRegex)
                 .Unless(request => request.Postcode == null && request.Radius == null);
             RuleFor(request => request.TypeId)
                 .Must(id => TypeIds().Contains(id.ToString()))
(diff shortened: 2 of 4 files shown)


GetIntoTeachingApi/Models/Location.cs
```diff
@@ -9,8 +9,13 @@ namespace GetIntoTeachingApi.Models
 {
     public class Location
     {
+        public static readonly Regex AreaOrFullPostcodeRegex = new Regex(
+            $"{AreaPostcodePattern}|{FullPostcodePattern}", RegexOptions.IgnoreCase);
         public static readonly Regex PostcodeRegex = new Regex(
-            @"^([A-Z][A-HJ-Y]?\d[A-Z\d]? ?\d[A-Z]{2}|GIR ?0A{2})$", RegexOptions.IgnoreCase);
+            FullPostcodePattern, RegexOptions.IgnoreCase);
+
+        private const string AreaPostcodePattern = @"^([A-Z][A-HJ-Y]?\d[A-Z\d]?)$";
```
ethax-ross, line 17: Can we rename this to `OutwardPostcodePattern`? a postcode "area" refers to the first one or two characters, not including the digits so it's a little misleading. 

It may also be worth having your private constants as `OutwardPostcodePattern` and `InwardPostcodePattern` then combining them in the public constants `OutwardOrFullPostcodeRegex` and `PostcodeRegex` as a way of DRYing up the outward pattern (or interpolate the outward pattern into the full pattern const and leave it as you have it).

ethax-ross (commented):
Looks good - if you want to belt and braces it then adding another test at the `Store` level to ensure the `SearchTeachingEvents` function behaves as expected with a partial postcode would be a good addition. Similar to:

https://github.com/DFE-Digital/get-into-teaching-api/blob/master/GetIntoTeachingApiTests/Services/StoreTests.cs#L390-L401

It's unlikely that it would fail here though (unless an additional form of postcode validation/sanitisation was added in this layer) so I'll leave it up to you if you want to add it in or not.

GetIntoTeachingApi/Models/Location.cs
```diff
@@ -9,8 +9,14 @@ namespace GetIntoTeachingApi.Models
...
     {
+        public static readonly Regex OutwardOrFullPostcodeRegex = new Regex(
+            $"^({OutwardPostcodePattern}|{FullPostcodePattern})$", RegexOptions.IgnoreCase);
         public static readonly Regex PostcodeRegex = new Regex(
-            @"^([A-Z][A-HJ-Y]?\d[A-Z\d]? ?\d[A-Z]{2}|GIR ?0A{2})$", RegexOptions.IgnoreCase);
+            $"^({FullPostcodePattern})$", RegexOptions.IgnoreCase);
+
+        private const string OutwardPostcodePattern = @"[A-Z][A-HJ-Y]?\d[A-Z\d]?";
+        private const string InwardPostcodePattern = @"\d[A-Z]{2}|GIR ?0A{2}";
+        private const string FullPostcodePattern = OutwardPostcodePattern + " ?" + InwardPostcodePattern;
```
jkempster34, line 20: Not sure if I have made this too convoluted, with the carets, dollars, parenthesis, and or pipe in the Regex object

GetIntoTeachingApi/Models/Location.cs
```diff
@@ -9,8 +9,14 @@ namespace GetIntoTeachingApi.Models
 {
     public class Location
     {
+        public static readonly Regex OutwardOrFullPostcodeRegex = new Regex(
+            $"^({OutwardPostcodePattern}|{FullPostcodePattern})$", RegexOptions.IgnoreCase);
         public static readonly Regex PostcodeRegex = new Regex(
-            @"^([A-Z][A-HJ-Y]?\d[A-Z\d]? ?\d[A-Z]{2}|GIR ?0A{2})$", RegexOptions.IgnoreCase);
+            $"^({FullPostcodePattern})$", RegexOptions.IgnoreCase);
+
```
jkempster34, line 16: Not sure if I have made this too convoluted, with the carets, dollars, parenthesis, and or pipe in the Regex object, but the rest in the const strings

ethax-ross: I think it's fine - I would maybe opt for interpolation in your `FullPostcodePattern`, but just my preference:

```
$"{OutwardPostcodePatter} ?{InwardPostcodePattern}"
```

ethax-ross (approved):
LGTM

GetIntoTeachingApi/Models/Location.cs
jkempster34: Ah, yeah it won't allow that because they are const. I can make them static readonly, I think that might work

PR #381, opened 2020-11-24
Use dotenv.net for development secrets

ethax-ross:
Previously we had to manually edit the `launchSettings.json` file with the development credentials for the CRM etc. This was susceptible to being accidentally committed into git (as we can't `gitignore` the `launchSettings.json` file completely).

Instead, this commit introduces `dotnet.env` for managing development secrets in an ignored `.env.development` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ msbuild.wrn
 
 # Coverage
 TestResults
+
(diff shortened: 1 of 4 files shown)


jkempster34 (approved):
Tested this and it all works as expected

PR #383, opened 2020-11-27
Default education phase to secondary for returners

ethax-ross:
We don't ask returners for their preferred education phase as part of the TTA sign up, instead defaulting it in the client-side to secondary. This is not ideal in the scenario where the candidate goes down the returner path then backs out and goes down the new candidate path; we end up pre-selecting secondary for their education phase even though they didn't explicitly select it.

Defaulting it in the API avoids this issue (we no longer have to set it in the client).


peteryates (approved):
Makes sense

PR #384, opened 2020-11-27
Adjust validation to allow null preferred education phase

ethax-ross:
If the candidate is returning to teaching we default the education phase so it doesn't need to be specified; this relaxes the validation rule.


peteryates (approved):
Makes sense

PR #386, opened 2020-11-30
Added labeller

sfawcett123:
When A PR is created the labeler will attempt to assign it a label based upon what code has changed.  
This is useful in the DevOps environment as we are monitoring repositories as a team, identifying work the team could do if an individual is not available

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,18 @@
+---
+
+VisualCpp:
+  - GetIntoTeachingApi/**/*.rb
+
+Test:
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,18 @@
+---
+
+VisualCpp:
+  - GetIntoTeachingApi/**/*.rb
```
ethax-ross, line 4: This should be `- GetIntoTeachingApi/**/*.cs` I think?

ethax-ross (approved):

PR #388, opened 2020-12-01
Use fallback booking quotas when CRM is unavailable

ethax-ross:
We want candidates to be able to sign up for a TTA even when the CRM is offline. Currently, there is a hard dependency on the CRM for the callback booking quotas as we query these at the point of displaying them to the user (so as to have the most up-to-date information).

The issue with this is that if the CRM is offline we will display an error to the user. Instead, we want to let the user proceed and book from a pre-generated list of fallback quotas.

The fallback quotas should include the next 5 working days (not including the current day) and have times from 9am to 4:30pm available in 30 minute intervals.

If the CRM is unavailable or returns an empty set of quotas, then we want to instead display the generated list of fallback options.


jebw (approved):
LGTM

PR #394, opened 2020-12-04
Add source to csv locations

jkempster34:
Some locations from Google Geocoding were being set in the database with the wrong SRID. This PR adds a source to the locations that are imported from FreeMapTools, and an endpoint to manually trigger a location batch job.

### DIFF
--- a/GetIntoTeachingApi/Controllers/OperationsController.cs
+++ b/GetIntoTeachingApi/Controllers/OperationsController.cs
@@ -2,9 +2,11 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using GetIntoTeachingApi.Attributes;
+using GetIntoTeachingApi.Jobs;
 using GetIntoTeachingApi.Models;
 using GetIntoTeachingApi.Services;
 using GetIntoTeachingApi.Utils;
+using Hangfire;
 using Microsoft.AspNetCore.Mvc;
 using Swashbuckle.AspNetCore.Annotations;
 
@@ -87,6 +89,18 @@ namespace GetIntoTeachingApi.Controllers
             System.Text.StringBuilder builder = null;
 
             builder.Append("throw error");
+        }
+
+        [HttpGet]
+        [Route("trigger_location_sync")]
+        [SwaggerOperation(
+            Summary = "Manually triggers a location sync job",
+            OperationId = "TriggerLocationSync",
+            Tags = new[] { "Operations" })]
+        [ProducesResponseType(typeof(void), 200)]
+        public void TriggerLocationSync()
+        {
+            RecurringJob.Trigger(JobConfiguration.LocationSyncJobId);
         }
     }
 }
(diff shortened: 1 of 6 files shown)


GetIntoTeachingApi/Controllers/OperationsController.cs
```diff
@@ -88,5 +91,18 @@ public void SimulateError()
...
+        [Authorize]
+        [Route("trigger_location_sync")]
+        [SwaggerOperation(
+            Summary = "Manually triggers a location sync job",
+            OperationId = "TriggerLocationSync",
+            Tags = new[] { "Operations" })]
+        [ProducesResponseType(typeof(void), 200)]
+        public void TriggerLocationSync()
+        {
+            RecurringJob.Trigger(JobConfiguration.LocationSyncJobId);
```
ethax-ross, line 105: Can we add a test for this in `OperationsControllerTests` and it would also be worth testing that the `Authorize` attribute is present, similar to:

https://github.com/DFE-Digital/get-into-teaching-api/blob/master/GetIntoTeachingApiTests/Controllers/CandidatesControllerTests.cs#L29-L33

GetIntoTeachingApi/Jobs/LocationBatchJob.cs
```diff
@@ -56,7 +56,7 @@ public async Task RunAsync(string batchJson)
 
         private static Location CreateLocation(dynamic location)
         {
-            return new Location(location.Postcode, location.Latitude, location.Longitude);
+            return new Location(location.Postcode, location.Latitude, location.Longitude, Source.CSV);
```
ethax-ross, line 59: Can we cover the source with a test in `LocationBatchJobTests` by putting something in here:

https://github.com/DFE-Digital/get-into-teaching-api/blob/master/GetIntoTeachingApiTests/Jobs/LocationBatchJobTests.cs#L49

Along the lines of:

```
var locations = DbContext.Locations.ToList();
locations.Count().Should().Be(batch.Count());
locations.All(l =>batch.Any(b => BatchLocationMatchesExistingLocation(b, l))).Should().BeTrue();
locations.All(l => l.Source == Soruce.Csv)
```

ethax-ross (commented):
Looks good, could just do with some test coverage 👍

GetIntoTeachingApiTests/Jobs/LocationBatchJobTests.cs
```diff
@@ -23,7 +24,7 @@ public class LocationBatchJobTests : DatabaseTests
         private readonly IMetricService _metrics;
         private readonly Mock<ILogger<LocationBatchJob>> _mockLogger;
```
jkempster34, line 26: Feel like I should have added a new test for the location source such as RunAsync_InsertsLocationsWithSourceAsCsv, but the shared fixture context is making that difficult, so is it okay to just be added to the existing test?

ethax-ross: Yep existing test is fine, the current test checks the postcode/lat/lng all get set correctly so I think it makes sense to add the check for the source in there as well

PR #395, opened 2020-12-04
Correct postcodes with unknown sources

jkempster34:
- Add endpoint to trigger location batch job.
- When a location batch job is run, replace locations where the source is unknown with CSV source. 
- Tags locations cached from Google Geocoding as "Google"
- Specifies "postcodes" in the Google Geocode request.

### DIFF
--- a/GetIntoTeachingApi/Services/Store.cs
+++ b/GetIntoTeachingApi/Services/Store.cs
@@ -270,7 +270,7 @@ namespace GetIntoTeachingApi.Services
 
         private async Task CacheLocation(string postcode, Point coordinate)
         {
-            var location = new Location(postcode, coordinate);
+            var location = new Location(postcode, coordinate, Source.Google);
             await _dbContext.Locations.AddAsync(location);
             await _dbContext.SaveChangesAsync();
         }
--- a/GetIntoTeachingApiTests/Controllers/OperationsControllerTests.cs
+++ b/GetIntoTeachingApiTests/Controllers/OperationsControllerTests.cs
@@ -12,6 +12,10 @@ using GetIntoTeachingApi.Attributes;
 using System;
 using System.Reflection;
 using Microsoft.AspNetCore.Authorization;
+using Hangfire;
+using Hangfire.Common;
+using GetIntoTeachingApi.Jobs;
+using Hangfire.States;
 
 namespace GetIntoTeachingApiTests.Controllers
 {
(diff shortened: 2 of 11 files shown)


GetIntoTeachingApiTests/Controllers/OperationsControllerTests.cs
```diff
@@ -103,5 +105,12 @@ public void LogRequests_IsPresent()
         {
             typeof(OperationsController).Should().BeDecoratedWith<LogRequestsAttribute>();
         }
+
+        [Fact]
+        public void TriggerLocationSync_Authorize_IsPresent()
```
ethax-ross, line 110: Can we add another test that calls the controller action and ensures the job is queued?

GetIntoTeachingApi/Services/Store.cs
```diff
@@ -270,7 +270,7 @@ private async Task<Point> CoordinateForPostcode(string postcode)
 
         private async Task CacheLocation(string postcode, Point coordinate)
         {
-            var location = new Location(postcode, coordinate);
+            var location = new Location(postcode, coordinate, Source.Google);
```
ethax-ross, line 273: This needs a test to cover the source being correctly set to `Google` when we're caching a geocoded response; can probably add it in here:

https://github.com/DFE-Digital/get-into-teaching-api/blob/master/GetIntoTeachingApiTests/Services/StoreTests.cs#L128

You can just change the last line:

```
DbContext.Locations.FirstOrDefault(l => l.Postcode == sanitizedPostcode && l.Source == Source.Google).Should().NotBeNull();
```

GetIntoTeachingApiTests/Jobs/LocationBatchJobTests.cs
```diff
@@ -45,9 +46,10 @@ public async void RunAsync_InsertsNewLocations()
             await _job.RunAsync(JsonConvert.SerializeObject(batch));
             await _job.RunAsync(JsonConvert.SerializeObject(batch));
 
-            DbContext.Locations.Count().Should().Be(batch.Count());
+            DbContext.Locations.Count().Should().Be(batch.Count);
             DbContext.Locations.ToList().All(l =>
                 batch.Any(b => BatchLocationMatchesExistingLocation(b, l))).Should().BeTrue();
+            DbContext.Locations.All(l => l.Source == Source.CSV);
```
ethax-ross, line 52: We've covered the scenario for when a new location gets the correct CSV source, but we're missing a test for the scenario where an existing location exists with Unknown source and is then updated to CSV - highlighted by SonarCloud here:

https://sonarcloud.io/component_measures?id=DFE-Digital_get-into-teaching-api&metric=new_coverage&pullRequest=395&selected=DFE-Digital_get-into-teaching-api%3AGetIntoTeachingApi%2FJobs%2FLocationBatchJob.cs&view=list

GetIntoTeachingApiTests/Controllers/OperationsControllerTests.cs
jkempster34: Had to rework this so that it could be tested properly. Got stuck for a while, turned out I was testing for job type of LocationBatchJob instead of LocationSyncJob.

ethax-ross (approved):

PR #396, opened 2020-12-04
Switch RateLimiting counters to be backed by Redis

ethax-ross:
Add `Microsoft.Extensions.Caching.StackExchangeRedis` package for connecting to a Redis instance.

Add Redis instance to docker-compose.

`RedisConfiguration` abstracts extraction of connection details from `VCAP_SERVICES` (which is also set in Development to point to the instance running in Docker).

Remove integration tests - I ran into flakiness here, it was passing in debug mode but failing when ran normally. It stems from the `WebApplicationFactory` instance used to boot up the app for the integration test; I believe we need to better stub out the CRM integration as an exception in a background thread is causing the test to fail when it attempts to call out to the CRM. Even when I thought I had all dependent services stubbed out it would still fail randomly with an unhelpful, generic error around dependency injection. These tests don't add a huge amount of value as they really only test the rate limiting configuration; the library has its own tests.

I believe in the Rails apps we prefix Redis keys as we use a shared Redis instance, but I don't think there is a way for me to do this in the API (the rate limiting library will determine the keys); that being said the chances of them clashing should be very small.

Also adds Redis to the health check operation.

There are no tests for the Redis health check as we would have to write an adapter to abstract the StackExchange Redis instance, which wouldn't leave us with much left to test so doesn't feel worth it right now. For some reason SonarCloud is reporting that `RedisConfiguration` is untested, which isn't right as it has full test coverage.

If Redis is unavailable for any reason an in-memory store will be used as a fallback option.


jebw (approved):
LGTM

jebw (approved):

PR #397, opened 2020-12-07
Add endpoint to remove locations of unknown source

jkempster34:
Some locations were being added to the database with incorrect coordinates, these have been tagged as having an unknown source. This PR adds and endpoint which removes them from the database.

### DIFF
--- a/GetIntoTeachingApi/Controllers/OperationsController.cs
+++ b/GetIntoTeachingApi/Controllers/OperationsController.cs
@@ -107,5 +107,25 @@ namespace GetIntoTeachingApi.Controllers
         {
             _jobClient.Enqueue<LocationSyncJob>(job => job.RunAsync(LocationSyncJob.FreeMapToolsUrl));
         }
+
+        [HttpGet]
+        [Authorize]
+        [Route("remove_unknown_locations")]
+        [SwaggerOperation(
+      Summary = "Remove locations of unknown source from the database",
+      OperationId = "RemoveUnknownLocations",
+      Tags = new[] { "Operations" })]
+        [ProducesResponseType(typeof(void), 200)]
+        [ProducesResponseType(typeof(string), 200)]
+        public async Task<IActionResult> RemoveUnknownLocations(bool dryRun)
+        {
+            if (dryRun)
+            {
+                return Ok($"Number of locations with Unknown source: {_store.GetNumberOfUnknownLocations()}");
+            }
+
+            await _store.RemoveUnknownLocations();
+            return Ok();
+        }
     }
 }
--- a/GetIntoTeachingApi/Controllers/OperationsController.cs
+++ b/GetIntoTeachingApi/Controllers/OperationsController.cs
@@ -115,7 +115,7 @@ namespace GetIntoTeachingApi.Controllers
       Summary = "Remove locations of unknown source from the database",
       OperationId = "RemoveUnknownLocations",
       Tags = new[] { "Operations" })]
-        [ProducesResponseType(typeof(void), 200)]
+        [ProducesResponseType(204)]
         [ProducesResponseType(typeof(string), 200)]
         public async Task<IActionResult> RemoveUnknownLocations(bool dryRun)
         {
@@ -125,7 +125,7 @@ namespace GetIntoTeachingApi.Controllers
             }
 
             await _store.RemoveUnknownLocations();
-            return Ok();
+            return NoContent();
         }
     }
 }
--- a/GetIntoTeachingApi/Services/Store.cs
+++ b/GetIntoTeachingApi/Services/Store.cs
@@ -120,14 +120,12 @@ namespace GetIntoTeachingApi.Services
 
         public int GetNumberOfUnknownLocations()
         {
-            IQueryable<Location> unknownLocations = GetUnknownLocations();
-            return unknownLocations.Count();
+            return GetUnknownLocations().Count();
         }
 
         public async Task RemoveUnknownLocations()
         {
-            IQueryable<Location> unknownLocations = GetUnknownLocations();
-            _dbContext.RemoveRange(unknownLocations);
+            _dbContext.RemoveRange(GetUnknownLocations());
             await _dbContext.SaveChangesAsync();
         }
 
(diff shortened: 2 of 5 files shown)


GetIntoTeachingApi/Controllers/OperationsController.cs
```diff
@@ -107,5 +107,25 @@ public void TriggerLocationSync()
...
+        [ProducesResponseType(typeof(string), 200)]
+        public async Task<IActionResult> RemoveUnknownLocations(bool dryRun)
+        {
+            if (dryRun)
+            {
+                return Ok($"Number of locations with Unknown source: {_store.GetNumberOfUnknownLocations()}");
+            }
+
+            await _store.RemoveUnknownLocations();
+            return Ok();
```
ethax-ross, line 128: You can return `return NoContent();` here which is ` 204`, so you could them have:

```
[ProducesResponseType(typeof(string), 200)]
[ProducesResponseType(204)]
```

^ the `typeof(void)` is unnecessary

GetIntoTeachingApi/Services/Store.cs
```diff
@@ -118,6 +118,24 @@ public IQueryable<TeachingEvent> GetUpcomingTeachingEvents()
                 .Where(e => e.StartAt >= DateTime.UtcNow);
         }
 
+        public int GetNumberOfUnknownLocations()
+        {
+            IQueryable<Location> unknownLocations = GetUnknownLocations();
```
ethax-ross, line 123: Not a big deal but I tend to favour using `var` where possible as it makes refactoring easier:

```
var unknownLocations = GetUnknownLocations();
```

This would be fine as a one-liner as well:

```
GetUnknownLocations().Count()
```

GetIntoTeachingApiTests/Controllers/OperationsControllerTests.cs
```diff
@@ -130,5 +131,56 @@ public void TriggerLocationSync_EnqueuesLocationSyncJob()
                                   (job.Method.Name == "RunAsync")),
                 It.IsAny<EnqueuedState>()));
         }
+
+        [Fact]
+        public void RemoveUnknownLocations_Authorize_IsPresent()
+        {
+            MethodInfo triggerLocationSync = typeof(OperationsController).GetMethod("RemoveUnknownLocations");
```
ethax-ross, line 138: Copy/paste error in variable name here

GetIntoTeachingApiTests/Controllers/OperationsControllerTests.cs
```diff
@@ -130,5 +131,56 @@ public void TriggerLocationSync_EnqueuesLocationSyncJob()
...
+
+        [Fact]
+        public async Task RemoveUnknownLocations_WhenDryRunSpecified_ReturnsMessage()
+        {
+            const bool dryRun = true;
+            const int mockNumberOfLocations = 10;
+            var mockStore = new Mock<IStore>();
+            mockStore.Setup(store => store.GetNumberOfUnknownLocations())
+                .Returns(mockNumberOfLocations);
+            var controller = new OperationsController(
```
ethax-ross, line 150: All this setup is already done for you in the constructor, this test case could just be:

```
public async Task RemoveUnknownLocations_WhenDryRunSpecified_ReturnsMessage()
{
  _mockStore.Setup(store => store.GetNumberOfUnknownLocations())
                .Returns(10);

  var response = await controller.RemoveUnknownLocations(true);
  
  var response = response.Should().BeOfType<OkObjectResult>() as ObjectResult;
  response.Value.Should().Be("Number of locations with Unknown source: 10");
}
```

I would also add a `Verify` on the store to ensure the `RemoveUnknownLocations` is never called

GetIntoTeachingApiTests/Controllers/OperationsControllerTests.cs
```diff
@@ -130,5 +131,56 @@ public void TriggerLocationSync_EnqueuesLocationSyncJob()
...
+        }
+
+        [Fact]
+        public async Task RemoveUnknownLocations_WhenDryRunIsFalse_CallsRemoveUnknownLocations()
+        {
+            const bool dryRun = false;
+            var mockStore = new Mock<IStore>();
+            mockStore.Setup(store => store.RemoveUnknownLocations())
+                .Verifiable();
+            var controller = new OperationsController(
```
ethax-ross, line 172: Ditto for test setup

ethax-ross (commented):
Looks good overall, just a few suggestions

GetIntoTeachingApiTests/Controllers/OperationsControllerTests.cs
jkempster34: I did this in a separate test. I don't know how important "one thing per test" is when you write the tests after doing the thing. But, I figured in this case they were separate enough to warrant a new test?

ethax-ross: I would typically group assertions that are very closely related into the same test case, mainly to avoid repeated setup/teardown logic, but since that's not a big issue here I'm not fussed either way 👍

GetIntoTeachingApiTests/Services/StoreTests.cs
```diff
@@ -495,6 +495,36 @@ public async void GetUpcomingTeachingEventsByTypeAsync_ReturnsUpcomingEventsGrou
...
+            var locations = new List<Location>() {
+                new Location { Postcode =  "ky119yu", Source = Source.Unknown },
+                new Location { Postcode = "ca48le", Source = Source.CSV },
+            };
+            DbContext.Locations.AddRange(locations);
+            await DbContext.SaveChangesAsync();
+
+            await _store.RemoveUnknownLocations();
+
+            DbContext.Locations.Count().Should().Be(1);
```
ethax-ross, line 525: Sorry I missed this on my first review, but it would be better to have an expectation that none of the `DbContext.Locations` have `Source.Unknown` - you could update the code to remove all `CSV` type locations and this test would still pass.

```
DbContext.Locations.All(l => l.Source != Source.Unknown).Should().BeTrue();
```

ethax-ross (approved):

PR #403, opened 2020-12-08
Fallback to in-memory store if Redis is unavailable

ethax-ross:
When using the built-in Redis store if the Redis instance becomes unavailable the app is effectively taken offline. Instead, this commit implements a custom `IRateLimitCounterStore` that will attempt to write to Redis and, if unavailable, it will fallback to an in-memory store.

I've tested this thoroughly locally by running it against the Redis instance in Postgres and killing it off, it seems to behave correctly (testing it would be difficult due to it being so closely to the library, which isn't open for mocking in its current state).


ethax-ross:
Closing - cherry picked this commit into the original PR.

PR #405, opened 2020-12-09
Devops/workflows/automate

sfawcett123:
### Automate workflow
Automate the API workflow through to QA.

### Routes
**Push** -> Build
                -> Sonarcloud
                -> .Net Test
                -> CVE

**PR Merge** -> Build -> Deploy Dev -> Deploy QA -> Cypress Test -> STOP
     (master)     -> Sonarcloud
                         -> CVE

**Manual**  -> Deploy Production


.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
...
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PERSONAL_ACCESS_TOKEN }}
+
+      - name: Build
+        uses: docker/build-push-action@v2
+        with:
+          push: ${{ steps.vars.outputs.push }}
```
vigneshmsft, line 84:
```suggestion
          push: ${{ github.ref == 'refs/heads/master' }}
```
would just adding the condition here work?

.github/workflows/build.yml
```diff
@@ -0,0 +1,271 @@
...
+      - name: Get Short SHA
+        id: vars
+        run: |
+             echo ::set-output name=sha_short::$(echo $GITHUB_SHA | cut -c -7)
+             if [ ${{github.ref}} == 'refs/heads/master' ]
+             then
+             echo ::set-output name=push::true
+             else
+             echo ::set-output name=push::false
+             fi
```
vigneshmsft, line 65:
```suggestion
             echo ::set-output name=push::${{ github.ref == 'refs/heads/master' }}
```
will this work?

.github/workflows/build.yml
sfawcett123: it might, but I think it might also be confusing when reading back on it in 100 years time ..

.github/workflows/build.yml
```diff
@@ -0,0 +1,266 @@
...
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@master
+
+      - name: Get Short SHA
+        id: vars
+        run: |
+             echo ::set-output name=sha_short::$(echo $GITHUB_SHA | cut -c -7)
+             echo ::set-output name=push::${{ github.ref == 'refs/heads/master' }}
```
vigneshmsft, line 60: we can now remove this?

.github/workflows/build.yml
```diff
@@ -0,0 +1,266 @@
...
+         with:
+           workflow: owasp
+           token: ${{ secrets.ACTIONS_API_ACCESS_TOKEN }}
+           inputs: '{"environment": "dev"}'
+
+       - name: Generate Tag from PR Number
+         id: tag_version
+         run: |
+           MATCH="Merge pull request #[0-9]+"
+           STRING=$(curl -s -X GET "https://api.github.com/repos/${{github.repository}}/commits/${{github.sha}}" | jq -r  ".commit.message")
```
vigneshmsft, line 139: I think instead of curl, you can use
${{ github.event.push.commits[0].message }}

.github/workflows/deploy.yml
```diff
@@ -1,36 +1,123 @@
...
-            continue-after-seconds: 180
+              terraform_version: 0.13.4
+
+       - name: Wait for any previous runs to complete
+         uses: softprops/turnstyle@v1
          env:
-           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+         with:
+           continue-after-seconds: 180
```
saliceti, line 120: Isn't it dangerous? We should fail after a long timeout, like 10-15min. And never continue.

.github/workflows/build.yml
sfawcett123: Agreed removed

.github/workflows/build.yml
sfawcett123: although it might work, I am going to leave it as is now knowing my bit works and add this as a future change

.github/workflows/deploy.yml
sfawcett123: actually that one can be removed there is a turnstyle at the begining

saliceti (approved):

vigneshmsft (dismissed):

vigneshmsft (approved):

ethax-ross (approved):

PR #406, opened 2020-12-09
Log request details on validation failure

ethax-ross:
Currently when the API receives a request the appropriate request model validator is ran automatically before the controller action is called. The side effect of this is that the `LoggableAttribute` action filter is never ran for these requests, meaning we don't have meaningful logs around what was in the request payload.

This PR breaks out the logic to translate an `ActionExecutingContext` and `ControllerActionDescriptor` into formatted, sanitised log messages to a new utility class `LoggableMessageComposer`.

We then hook into the `PostConfigure` callback and override the `InvalidModelStateResponseFactory` to log out the request details before calling the built in factory that formats and returns a helpful 400 response..

### DIFF
--- a/GetIntoTeachingApi/Utils/LoggableMessageComposer.cs
+++ b/GetIntoTeachingApi/Utils/LoggableMessageComposer.cs
@@ -1,0 +1,60 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using GetIntoTeachingApi.Attributes;
+using Microsoft.AspNetCore.Mvc.Controllers;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace GetIntoTeachingApi.Utils
+{
+    public class LoggableMessageComposer
+    {
+        public static string LogMessage(string prefix, ControllerActionDescriptor descriptor, IEnumerable<string> messages)
+        {
+            var actionName = descriptor.ActionName;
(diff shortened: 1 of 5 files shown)


GetIntoTeachingApi/Utils/LoggableMessageComposer.cs
```diff
@@ -0,0 +1,60 @@
...
+            var payload = string.Join("\n", messages.Where(m => !string.IsNullOrEmpty(m)));
+
+            return $"{prefix} {controllerName}:{actionName} - {payload}";
+        }
+
+        public static string LogMessageForObject(object obj)
+        {
+            var type = obj.GetType().GetGenericArguments().FirstOrDefault();
+
+            if (type == null)
```
peteryates, line 26: If this is set to `.FirstOrDefault()` just above will it ever be `null`?

peteryates (approved):
Looks great, one minor question

GetIntoTeachingApi/Utils/LoggableMessageComposer.cs
ethax-ross: Yeah, so if we did `.First()` here it would throw an exception if there were no elements, but `FirstOrDefault()` really means first or `null` if there are no elements. Not very clear naming by LINQ IMO...

peteryates: Ah, makes sense. Thanks 👍🏽

PR #409, opened 2020-12-09
Only run CrmSyncJob on first boot

ethax-ross:
We used to run this every time the app booted as syncs were infrequent and it could have been helpful, but we now run the CrmSync every 5 minutes so its not really necessary any longer.

Also this may prevent the exception thats often raised when the app boots and tries to sync before the permissions to the DB have propagated.


peteryates (approved):
Makes sense

PR #413, opened 2020-12-10
Revert "Enable rate limiting in test environment"

ethax-ross:
Reverts DFE-Digital/get-into-teaching-api#412 - this was enabled temporarily in the test environment so that we can ensure the rate limiting counters work correctly when backed by the Redis store.


peteryates (approved):
👍🏽

PR #418, opened 2020-12-11
Optimize location sync job

ethax-ross:
The location sync operation takes around 16 minutes in development, which can be a bit of a pain and is also a risk if we had to re-run the sync with urgency in production for any reason. This PR adds optimisations that bring that time down to just over 2 minutes.

**Only hard code number of workers in hosted environments**

By default Hangfire will use `5 * the number of cores the instance has` as the number of workers; in production this was resulting in >100 workers per instance and draining our DB connection pool so we hard coded the count to be 20. In development we only run on a single machine that likely has fewer cores so using the Hangfire default worker count gives us a little better performance.

**Optimize postcocde sanitization**

We were using a `Regex` replace here which is a concise but inefficient way of removing white space from a string. As this method is called once per row in the CSV its ran around 1.7 million times per sync, so doing a micro-optimization on the sanitisation offers a performance boost.

**Optimize LocationSyncJob**

The `LocationSyncJob` and associated `LocationBatchJob` were taking around 16 minutes to complete when ran locally. This commit improves the performance to just over 2 minutes by:

- Condensing the work into a single long running job
- Taking advantage of Postgres `INSERT INTO ... ON CONFLICT` for bulk inserting
- Tweaking the insert batch size to 300

Whilst this brings a large performance gain it has a couple of disadvantages, though IMO they are not going to be a huge issue for us:

- If the job fails we can't just re-run a small batch of failures, we have to do the whole lot
- We're coupling the logic to Postgres

**Remove `LocationBatchJob` monitoring panel**

This job no longer exists, so we can remove the panel in Grafana.


GetIntoTeachingApi/Jobs/LocationSyncJob.cs
```diff
@@ -99,28 +101,41 @@ private async Task SyncLocations(string csvPath)
...
-
-            // Batch is serialized to pass by value.
-            _jobClient.Enqueue<LocationBatchJob>(x => x.RunAsync(JsonConvert.SerializeObject(batch)));
+            }
+
+            var parameters = batch.SelectMany((location, index) => DbParameters(location, index));
+            var values = string.Join(", ", batch.Select((_, index) => $"(@postcode{index}, @coordinate{index}, @source{index})"));
+
+            await _dbContext.Database.ExecuteSqlRawAsync(
+                $"INSERT INTO public.\"Locations\" (\"Postcode\", \"Coordinate\", \"Source\") VALUES {values} ON CONFLICT DO NOTHING", parameters);
```
peteryates, line 124: I guess this is using raw SQL rather than something like `SqlBulkCopy` because it's PostgreSQL rather than SQL Server?

peteryates (approved):
This looks great.

GetIntoTeachingApi/Jobs/LocationSyncJob.cs
ethax-ross: I thought about using `COPY` as that would be lightning fast, but the issue is we can't straight ingest the CSV as we sanitize the postcode (removing white space and lower casing) and convert the lat/lng to a `Point` with the correct SRID.

PR #420, opened 2020-12-11
Split TypeEntity into a new LookupItem model

ethax-ross:
- Rename type-related methods to avoid clashes …

We want to use `LookupItem` as the new model name, but we already use this as the method name when querying various type-related models; refactoring to free up the term going forward.

- Split LookupItems out of TypeEntities

Currently we have two slightly different objects being returned as `TypeEntity` models; they each have an `Id` and `Value`, however they differ on the type of their `Id` attributes (one being `Integer` and the other a `Guid`). Originally I decided to return them all as part of the `/types` endpoints and changed the `Id` to be `String` to accommodate both.

The issue with doing this is that the API is inconsistent; it will return an `Integer` as a `String` for the type `Id` but in other endpoints where you use this `Id` as a reference it is always expected to be an `Integer`. This involves casting on the client-side, which is less than ideal.

Instead, this commit initially breaks out the types with a `Guid` `Id` into a new `LookupItem` class. The `/types` endpoints remain unchanged for now for backwards compatibility.

Going forward we should split out the `Integer` `Id` types into another model, update the clients to use the new endpoints and then bin all the type-related code.

- Migration to add LookupItem model


jebw (approved):
LGTM

PR #421, opened 2020-12-11
Inject ICrmService into the Store

ethax-ross:
- Inject ICrmService into the Store

We pass an instance of `ICrmService` into various methods of the `IStore`. This makes the implementation more verbose than it needs to be; injecting an instance directly into the `IStore` instead of specific methods simplifies things.


jebw (approved):
LGTM

PR #425, opened 2020-12-16
Slack Deploy and Notifiy on Release

sfawcett123:
### Publish Release on SLACK

- When the API is ready for live publish the release information on slack
- Run the deployment process to push the API to LIVE


.github/workflows/build.yml
```diff
@@ -210,22 +210,56 @@ jobs:
...
          env:
-           SLACK_USERNAME: GiT Deployment
-           SLACK_CHANNEL: getintoteaching_tech
-           SLACK_COLOR: '#B90E0A'
-           SLACK_ICON: https://github.com/rtCamp.png?size=48
-           SLACK_TITLE: Production Release ${{github.event.title}}
-           SLACK_MESSAGE: Failure deploying Production release
-           SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
-           SLACK_FOOTER: ${{env.APPLICATION}}
+           SLACK_COLOR:    env.SLACK_FAILURE
```
jebw, line 260: Should these be `${{ env.SLACK_FAILURE }}` and `${{ env.SLACK_ICON }}`

.github/workflows/build.yml
```diff
@@ -31,9 +34,8 @@ jobs:
         if: failure()
         uses: rtCamp/action-slack-notify@master
         env:
-           SLACK_CHANNEL: getintoteaching_tech
```
jebw, line 34: How does this find which channel to publish to? is the `getintoteaching_tech` the default one?

.github/workflows/build.yml
```diff
@@ -87,9 +89,8 @@ jobs:
...
-           SLACK_COLOR: '#3278BD'
-           SLACK_ICON: https://github.com/rtCamp.png?size=48
+           SLACK_COLOR: env.SLACK_FAILURE
```
jebw, line 92: Does this  need the brace syntax?

.github/workflows/build.yml
```diff
@@ -12,6 +12,9 @@ env:
   DOMAIN:                london.cloudapps.digital
   APPLICATION:           Get Into Teaching API Service
   PAAS_APPLICATION_NAME: get-into-teaching-api
+  SLACK_FAILURE:         '#ff0000'
```
jebw, line 15: I think we need an 'alert' colour here

eg 

Green = success
Red = Failed deploy
Amber = Testsuite failure on branch

jebw (commented):
I think we've a question of how to handle notification of failures on PRs

saliceti (dismissed):

saliceti (approved):

.github/common_environment.yml
```diff
@@ -0,0 +1,9 @@
+  DOCKERHUB_REPOSITORY:  dfedigital/get-into-teaching-api
+  DOMAIN:                london.cloudapps.digital
+  APPLICATION:           Get Into Teaching API Service
+  PAAS_APPLICATION_NAME: get-into-teaching-api
+  SLACK_FAILURE:         '#ff0000'
+  SLACK_SUCCESS:         '#00ff00'
+  SLACK_ICON:            https://raw.githubusercontent.com//DFE-Digital/get-into-teaching-api/master/.github/image.png?size=48
```
jebw, line 7: Double slash after the domain doesn't look right?

.github/common_environment.yml
```diff
@@ -0,0 +1,9 @@
...
+  SLACK_USERNAME:        GiT Workflows
+  SLACK_FOOTER:          Get Into Teaching API Service
```
jebw, line 9: Would it make sense (and is it possible) - to use `${{ env.APPLICATION }}` here - it seems to duplicate that value, means if the app name gets changed, the footer is automatically updated?

.github/workflows/sonarcloud.yml
```diff
@@ -9,14 +9,17 @@ jobs:
...
-      - uses: actions/setup-java@v1
-        with:
-          java-version: '11'
-
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
 
+      - name: set-up-environment
+        uses: DFE-Digital/github-actions/set-up-environment@master
```
jebw, line 17: I'm somewhat wary of the `@master` - one bad commit to that repo will break builds for every dependant project in DFE?

Would it not be better to use tags and make the projects choose the versions they use? That being said we're using `@master` for the slack notifier and again, a bad change will break all our repos so maybe its not really a problem?

jebw (changes requested):
Looks good, only change I think is needed is the double slash, the other two are more just my opinions/thoughts

.github/common_environment.yml
sfawcett123: it would but I don't think our code is advanced enough just yet

.github/workflows/sonarcloud.yml
sfawcett123: its our team that manages this via PRs and we are reviewing and testing, so Master should be protected

jebw (dismissed):
LGTM

PR #426, opened 2020-12-17
Dry up validation with custom property validator

ethax-ross:
- Add custom property validators for pick list/lookup item ids

We often validate a foreign key against the available pick list/lookup items that are in the CRM. Until know we've been doing this in a fairly verbose way; instead this adds a custom property validator that will make the syntax more concise.

- Update validation rules to use PickList/LookupItemIdValidator

Updates the validation rules that were previously querying the store in private methods to now use a custom property validator that handles querying the store and matching an `Id`.

### DIFF
--- a/GetIntoTeachingApiTests/Validators/LookupItemIdValidatorTests.cs
+++ b/GetIntoTeachingApiTests/Validators/LookupItemIdValidatorTests.cs
@@ -33,7 +33,7 @@ namespace GetIntoTeachingApiTests.Validators
         {
             var selector = ValidatorOptions.ValidatorSelectors.DefaultValidatorSelectorFactory();
             var validationContext = new ValidationContext(_item.Id, new PropertyChain(), selector);
-            var propertyValidatorContext = new PropertyValidatorContext(validationContext, PropertyRule.Create<Guid, Guid>(t => t), "prop");
+            var propertyValidatorContext = new PropertyValidatorContext(validationContext, PropertyRule.Create<Guid, Guid>(t => t), "Prop");
 
             var errors = _validator.Validate(propertyValidatorContext);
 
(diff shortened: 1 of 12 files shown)


GetIntoTeachingApiTests/Validators/LookupItemIdValidatorTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        [Fact]
+        public void IsValid_WhenInvalidId_ReturnsFalse()
+        {
+            var selector = ValidatorOptions.ValidatorSelectors.DefaultValidatorSelectorFactory();
+            var validationContext = new ValidationContext(Guid.NewGuid(), new PropertyChain(), selector);
+            var propertyValidatorContext = new PropertyValidatorContext(validationContext, PropertyRule.Create<Guid, Guid>(t => t), "prop");
+
+            var errors = _validator.Validate(propertyValidatorContext);
+
+            errors.Should().NotBeEmpty();
```
jebw, line 52: Should this be checking the error messages?

GetIntoTeachingApiTests/Validators/PickListItemIdValidatorTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        [Fact]
+        public void IsValid_WhenInvalidId_ReturnsFalse()
+        {
+            var selector = ValidatorOptions.ValidatorSelectors.DefaultValidatorSelectorFactory();
+            var validationContext = new ValidationContext(456, new PropertyChain(), selector);
+            var propertyValidatorContext = new PropertyValidatorContext(validationContext, PropertyRule.Create<int?, int?>(t => t), "prop");
+
+            var errors = _validator.Validate(propertyValidatorContext);
+
+            errors.Should().NotBeEmpty();
```
jebw, line 51: Again should this be checking the error messages?

jebw (changes requested):

jebw (approved):

PR #428, opened 2020-12-21
Long lived candidate access tokens

ethax-ross:
- Add ability to generate long-lived access tokens for candidates

We want to be able to email candidates with a URL that links them straight into a pre-filled mailing list sign up form. In order to do this we need to be able to generate long-lived access tokens that are stored against a `contact` record in Dynamics. We can then send these out to candidates and use the token to retrieve their details from the API before presenting the form to them.

Add endpoint for generating long-lived access tokens against a candidate. Tokens are stored in Dynamics and are not returned, so the only way to expose them is via an email from the CRM itself.

- Exchange long lived access token for mailing list sign up

When a long lived access token is emailed out to a candidate we need a way for the Rails app to exchange this for the candidate information, so that the sign up form can be pre-filled.

Add method to exchange a long lived access token for mailing list sign up.

**Notes**

- Long lived tokens do not currently expire; we should store the date the token was generated so that we can verify it's still 'live'.
  - If we expire access tokens, we should auto-email the candidate with a newly generated token if they try and access using an expired one?
- Any API client could issue this request, but it would be ideal if only the CRM client could do this.
  - Add client API key for the CRM with elevated permissions.
- A GUID is used for the token; it may be worth using a more secure/random token generation?

### DIFF
--- a/GetIntoTeachingApi/Controllers/CandidatesController.cs
+++ b/GetIntoTeachingApi/Controllers/CandidatesController.cs
@@ -1,7 +1,10 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using GetIntoTeachingApi.Attributes;
+using GetIntoTeachingApi.Jobs;
 using GetIntoTeachingApi.Models;
 using GetIntoTeachingApi.Services;
+using Hangfire;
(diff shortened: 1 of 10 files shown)


ethax-ross:
This was picked up in a new series of PRs.

PR #430, opened 2020-12-23
Add TeachingEventsByType response model

ethax-ross:
Introduces a new response model for returning teaching events grouped by type. Previously we were using the format:

``` { "<typeid>": [<events>] } ```

This isn't a very RESTful response and also causes the type id to be returned in a string, where in actuality its an integer value.

This commit updates the response to be:

``` [{ "typeId": <typeid>, "teachingEvents": [<events>] }, { ... }] ```

In order to introduce the new response type an additional endpoint have been added:

``` 
teaching_events/search_grouped_by_type
```

These replace the existing methods, that will be removed once the clients have been updated (note the `upcoming_indexed_by_type` endpoint is going to be deprecated, which is why there is no equivalent `_grouped_` endpoint above for it):

``` 
teaching_events/search_indexed_by_type
teaching_events/upcoming_indexed_by_type
```

### DIFF
--- a/GetIntoTeachingApiTests/Models/TeachingEventsByTypeTests.cs
+++ b/GetIntoTeachingApiTests/Models/TeachingEventsByTypeTests.cs
@@ -1,0 +1,6 @@
+﻿namespace GetIntoTeachingApiTests.Models
+{
+    public class TeachingEventsByTypeTests
+    {
+    }
+}
--- a/GetIntoTeachingApi/Controllers/TeachingEventsController.cs
+++ b/GetIntoTeachingApi/Controllers/TeachingEventsController.cs
@@ -116,22 +116,6 @@ namespace GetIntoTeachingApi.Controllers
         [HttpGet]
         [CrmETag]
         [PrivateShortTermResponseCache]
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: GetIntoTeachingApiTests/Models/TeachingEventsByTypeTests.cs [hunk 1]
```diff
@@ -0,0 +1,6 @@
+﻿namespace GetIntoTeachingApiTests.Models
+{
+    public class TeachingEventsByTypeTests
+    {
```
jebw, line 4: Should this spec be empty?

jebw (commented):
Should the TeachingEventsByTypeTests be empty?

GetIntoTeachingApiTests/Models/TeachingEventsByTypeTests.cs
ethax-ross: Yeah the model only has getters/setters that are tested indirectly, no real value testing them explicitly but I like to have the test file there for if/when its needed.

jebw (approved):
LGTM

PR #431, opened 2021-01-04
Return all 'Online Events' on distance-based search

ethax-ross:
[Trello-660](https://trello.com/c/npt5sr4O/660-add-functionality-within-the-api-to-show-non-postcode-tagged-virtual-ttt-events-in-all-postcode-searches)

At the moment we have a few different kinds of 'Online Event':

- Events with type 'Online Event'
- Events with `IsOnline` set to `true` and no `Buillding.AddressPostcode`
- Events with `IsOnline` set to `true` and a `Building.AddressPostcode` (virtual)

When a radius-based search is performed we currently return events with type 'Online Event' in the results.

Going forward there will be another kind of fully online event type (not virtual) that is a 'Train to Teach' type event where `IsOnline` is `true` and there is no `Building.AddressPostcode`.

Instead of only returning 'Online Event' type events in radius-based searches, we want to return any event that is an online, but not 'virtual' event.

### DIFF
--- a/GetIntoTeachingApi/Models/TeachingEvent.cs
+++ b/GetIntoTeachingApi/Models/TeachingEvent.cs
@@ -68,6 +68,7 @@ namespace GetIntoTeachingApi.Models
         public TeachingEventBuilding Building { get; set; }
         [JsonIgnore]
         public Guid? BuildingId { get; set; }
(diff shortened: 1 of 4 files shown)


jebw (approved):
LGTM

PR #433, opened 2021-01-04
Change distance filter to work in database

jkempster34:
Geometry.Distance is translated to PostGIS [ST_Distance](https://postgis.net/docs/ST_Distance.html). Because teachingEvent.Building.Coordinate is attributed with column type geography, and because origin is implicitly converted to the geography type (it has an SRID), ST_Distance returns the distance between two points in meters.

This means that the events distance filtering can be performed in the database instead of in memory.


ethax-ross (approved):
💯 Nice!

ethax-ross (approved):

PR #434, opened 2021-01-05
Sync archived teaching events

ethax-ross:
[Trello-672](https://trello.com/c/Mx2EzMU4/672-development-create-an-archive-repository-for-online-events-which-are-3-4-months-old)

We are going to be adding a new page to the website that will allow users to browse 'archived' online events - these are events that are now in the past, up to 4 months old.

This commit updates the sync logic to query and cache events as far back as 4 months ago. The 4 month threshold isn't 'strict', so for ease I've treated it as 4 * 31 days.

The app needs to be updated first to ensure all requests are scoped to a date, so as not to inadvertently include archived events under 'all months'.


jebw (approved):
LGTM

PR #435, opened 2021-01-05
Allow event search with only start after date

ethax-ross:
Currently if you try and do an event search for all events after a date it throws a validation error. Instead, we should allow this query and return all events after the given dates.

Whilst this should have been possible anyway the use case was not required in the app, but going forward we need to be able to do this exclude archived events (being introduced shortly) from the event search.


peteryates (approved):
Looks good to me

PR #443, opened 2021-01-07
Add manual migration/rollback instructions for production

ethax-ross:
Originally I expected to be able to run `dotnet ef` CLI commands after ssh'ing into the production docker container via cloud foundry. Upon checking this in development it turns out the `dotnet ef` CLI won't function without the application source code being available, which isn't the case (in prod we only have the compiled DLL).

Instead, the only way to manually apply/rollback migrations in production involves generating the SQL queries locally, then creating a conduit into the production database and manually executing the SQL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,18 @@ The Hangfire web dashboard can be accessed at `/hangfire` in development.
 
 We run Entity Framework Core in order to persist models/data to a Postgres database.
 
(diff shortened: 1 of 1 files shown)


jebw (dismissed):

sfawcett123 (commented):
can I suggest you add link

https://docs.cloud.service.gov.uk/deploying_services/postgresql/#connect-to-a-postgresql-service-from-your-local-machine

sfawcett123 (approved):
Great

PR #448, opened 2021-01-11
Collect metrics for Timed One Time Passwords (TOTPs)

ethax-ross:
- Add metrics for TOTP generation/verification

We want to be able to track how many TOTPs are generated and, out of those, which are successfully verified (and, similarly, which are generated but never verified).

The metrics collect the `candidate_id` and the token so that we can attribute metrics to specific users without storing PII. The outcome of the verification process is also stored as a label, so we can determine if the correct TOTP was provided.

- Increment TOTP generated/verified metrics

When a TOTP is generated or verified we want to increment the corresponding metric. In order to link OTPs to a candidate the TOTP interfaces have been expanded to include a `candidateId`.

We also store the TOTP itself in the metric label so that we can see when the candidate verified the TOTP and _which_ one was verified (if they generated multiple). We store these in plain text as they are within our internal influxdb instance, only last 15 minutes and are useless without the candidate information that we don't store.

### DIFF
--- a/GetIntoTeachingApi/Controllers/CandidatesController.cs
+++ b/GetIntoTeachingApi/Controllers/CandidatesController.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using GetIntoTeachingApi.Attributes;
 using GetIntoTeachingApi.Models;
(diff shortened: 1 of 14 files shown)


jebw (approved):
LGTM

PR #454, opened 2021-01-15
Automatically run Fix network Policies

sfawcett123:
When deploying to Development run the Fix Network Policies after the run


ethax-ross (approved):
🎉

PR #456, opened 2021-01-15
Add TOTP metrics to engagement row

ethax-ross:
We are now sending metrics for generated/verified Timed One Time Passwords. This enables us to see how many we are generating vs how many are being verified, in addition to how many times a user is failing to enter a valid TOTP.

Add Grafana panels to visualize the new metrics.

Screenshot 2021-01-15 at 11 51 46


sfawcett123 (approved):
Looks good we need to deploy to Dev and then decide :)

PR #467, opened 2021-02-01
Add ApiClientHandler for authenticating api clients

ethax-ross:
[Trello-789](https://trello.com/c/h8Dxra2B/789-implement-way-to-manage-multiple-client-keys-in-the-api-with-varying-permissions)

This PR is the third/fourth step towards supporting multiple, independent clients with their own API keys and roles:

- [x] Add client modelling/validation
- [x] Load client definitions from YAML
- [x] Add new authentication handler
- [x] Scope controller/actions to roles
- [ ] Support rate limiting per client
- [ ] Update README
- [ ] Transition GiT/TTA to the new auth method

----

- Add ApiClientHandler for authenticating clients

Introduce an `ApiClientHandler`, which incorporates the existing `SharedSecret` logic. This will allow us to transition the existing services to the new format by:

```
- Adding an entry for them in clients.yml
- Adding the associated env variables
- Add an applicable role/scope access
- Updating the client services to use the new creds
```

We can then remove the deprecated `SharedSecret` logic from the API.

The `SharedSecret` code has also been updated to manually assign the `Admin` role, so that existing services retain the same access to all endpoints until we can migrate them to using the `ApiClient` authentication mode.

- Add development API key for admin client

Adding the env variable containing an admin API key that can be used in development. Until the `SharedSecretHandler` is removed we will have two ways to authenticate in development, via the legacy method using `abc123` or by the new handler using `secret`.

- Apply Admin role to existing controllers

This will grant clients authenticating with the legacy `SharedSecret` method permission to access all endpoints (which they currently have already) and the `Administrator` client will also have this permission using the new `ApiClientHandler`. Some controllers have the `Authorize` attribute without a role, which means any client regardless of role can access them.


peteryates (approved):
LGTM

PR #469, opened 2021-02-02
Enable rate limiting per API client service

ethax-ross:
[Trello-789](https://trello.com/c/h8Dxra2B/789-implement-way-to-manage-multiple-client-keys-in-the-api-with-varying-permissions)

This PR is the sixth step (support rate limiting per client) towards supporting multiple, independent clients with their own API keys and roles:

- [x] Add client modelling/validation
- [x] Load client definitions from YAML
- [x] Add new authentication handler
- [x] Scope controller/actions to roles
- [x] Transition GiT/TTA to the new auth method
- [x] Support rate limiting per client
- [ ] Update README

----

- Enable per-client rate limit rules

Add a custom `ResolveContributor` that looks up the API client from their API key and sets the `clientId` to the `ApiKeyPrefix`.

Define custom rate limiting rules for `GIT` and `TTA` clients. 

Add development api keys for `GIT`/`TTA` in addition to `ADMIN`.

The `ADMIN` client has been white-listed and a set of default rate limits imposed on potentially exploitable endpoints.

Bumped the `AspNetCoreRateLimit` library.

This PR doesn't have test coverage as I've hit an issue with the integration tests not loading in the client-specific policies for some reason even though they load in the development environment. Oddly the baseline rate limits load fine in the test environment. I've manually verified the client-level rate limits work correctly.

### DIFF
--- a/GetIntoTeachingApi/GetIntoTeachingApi.csproj
+++ b/GetIntoTeachingApi/GetIntoTeachingApi.csproj
@@ -5,7 +5,7 @@
 	</PropertyGroup>
 
 	<ItemGroup>
-		<PackageReference Include="AspNetCoreRateLimit" Version="3.0.5" />
+		<PackageReference Include="AspNetCoreRateLimit" Version="3.2.2" />
 		<PackageReference Include="CsvHelper" Version="15.0.5" />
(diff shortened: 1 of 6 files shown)


peteryates (approved):
Looks good, makes sense.

PR #470, opened 2021-02-02
Update subscriptions to not opt TTA candidates into post/bulk post

jkempster34:
Candidates signing up for TTA are currently opted into postal/bulk postal mail if returning to teaching. This is not longer wanted.


ethax-ross (approved):
LGTM

PR #471, opened 2021-02-02
Add GiT services as API clients

ethax-ross:
- Add GiT services as API clients

Adds the Get into Teaching services to the `clients.yml`. We can merge this and the two apps will continue to use the legacy secrets until we set up the new secrets in Azure and transition them over.

- Assign GiT API client roles

Assigns roles for the GiT API clients. This will limit access to only the controllers/actions required by the service.


peteryates (approved):
Neat

PR #472, opened 2021-02-02
 Initial Migration to Azure Secrets

sfawcett123:
The first phase is to deliver the terraform application using the
secrets, and make sure that works before rolling forward with the
workflows


ethax-ross (approved):
Looks good

PR #477, opened 2021-02-04
Add test coverage to client-level rate limits

ethax-ross:
- Update Startup to enable app to boot in test env

When trying to boot the application in the test environment there were a few things preventing the application from successfully starting (which we need to be able to do integration testing):

1. The `CdsServiceClient` is unable to connect (we now support running without the client, though it will crash if mocking is absent).
2. The database was not being configured for test (we can now connect to the test DB running in docker).
3. `Program` is not ran by `WebApplicationFactory` (code now shifted to be in the `Startup.Configure` instead, which gets run).

- Add WebApplicationFactory to run app in test env

The WebApplicationFactory allows us to boot an instance of the application running in the test environment and re-configure services that need to run differently for tests.

The only change we make currently is to run the rate limit counters against an in-memory store instead of Redis so that they are isolated between test runs.

- Support isolated integration tests against a Postgres DB

We already have the `DatabaseFixture` class to support unit testing against an isolated Postgres DB (running in Docker). This commit updates the `Environment` so that any integration tests using `DatabaseFixture` will also hit an isolated test database in docker.

- Add integration tests for rate limit behaviour

### DIFF
--- a/GetIntoTeachingApi/Adapters/CdsServiceClientWrapper.cs
+++ b/GetIntoTeachingApi/Adapters/CdsServiceClientWrapper.cs
@@ -10,8 +10,12 @@ namespace GetIntoTeachingApi.Adapters
 
         public CdsServiceClientWrapper(IEnv env)
         {
-            CdsServiceClient = new CdsServiceClient(ConnectionString(env));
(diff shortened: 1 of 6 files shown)


jebw (approved):
LGTM

PR #479, opened 2021-02-04
Refactor actions to exchange access tokens for candidate data

ethax-ross:

This PR deals with the first step of migrating existing candidates onto the new privacy policy via a mailing list sign up in the GiT app:

- [ ] Add new magic link token endpoints (creating tokens and exchanging for mailing list candidate data)
- [ ] Expire magic link tokens after 48 hours and ensure they are only single-use tokens. Write to field indicating when token is required/set.
- [ ] Add an API client for the CRM with access to the magic link generation endpoint

---

We currently only have one way of the API clients obtaining candidate data, which is by exchanging a user-generated access token for a resource (pre-filled mailing list/tta/event sign up):

```
POST /path/to/resource/{accessToken}
```

Going forward we are also going to support 'magic links' as a way of authenticating users. To that end, the endpoints are being refactored so we will ultimately be left with:

```
POST /path/to/resource/exchange_access_token/{accessToken}
POST /path/to/resource/exchange_magic_link_token/{magicLinkToken}
```

This PR adds in the `/path/to/resource/exchange_access_token/{accessToken}` endpoint and leaves the current endpoint in for backwards-compatibility. Once this is merged down the clients can be updated and the original endpoint removed.

### DIFF
--- a/GetIntoTeachingApi/Controllers/MailingListController.cs
+++ b/GetIntoTeachingApi/Controllers/MailingListController.cs
@@ -86,5 +86,31 @@ exchanged for your token matches the request payload here).",
 
             return Ok(new MailingListAddMember(candidate));
         }
+
+        [HttpPost]
+        [Route("members/exchange_access_token/{accessToken}")]
+        [SwaggerOperation(
+            Summary = "Retrieves a pre-populated MailingListAddMember for the candidate.",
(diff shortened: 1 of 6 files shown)


peteryates (approved):
Nice

PR #480, opened 2021-02-05
Initial support for magic link tokens

ethax-ross:

This PR deals with the second step of migrating existing candidates onto the new privacy policy via a mailing list sign up in the GiT app:

- [x] Add new magic link token endpoints (creating tokens and exchanging for mailing list candidate data)
- [ ] Expire magic link tokens after 48 hours and ensure they are only single-use tokens. Write to field indicating when token is generated/used.
- [ ] Add CRM client and restrict access to magic link generation endpoint. 
- [ ] Add job to process pending magic link generations. Connect models to CRM attributes (need CRM changes in prod first)

---

- Add magic link token fields to candidate

Add a field to store a magic link token against a candidate and the datetime that the token was created, so that we can enforce token expiry.

Only the token field itself is mapped to the CRM for now; the others will be mapped later on.

So that we can deploy this to production the `EntityFieldAttribute` has been updated with an `Transient` option. If set, the `BaseModel` will not map the fields to/from CRM `Entity` objects. This means we can safely deploy without the CRM having been updated to include the new field.

- Update CrmService to support magic links

We need to be able to query multiple candidates by their id from the CRM, so that we can bulk-process the magic link token generation (the CRM will be sending emails to thousands of candidates).

In addition, we need to be able to take a magic link from a request and find the corresponding candidate record. This method returns a collection of candidates in the edge case that a magic link has been duplicated during generation or when a contact record has been copied in the CRM. It has a null/empty string guard statement to ensure potentially expensive queries are avoided (most contacts _won't_ have a magic link token set).

- Add magic link token generator service

Add a service to generate secure, random 128 bit magic link tokens. It also has a method that will exchange a valid token for the associated `Candidate` record.

In the unlikely event that two candidates have been assigned the same magic link token, the service will return `null` (ultimately forcing a re-generation of the token for that candidate).

- Add endpoint to generate candidate magic link tokens

Introduce a new endpoint that will enable the CRM to batch-generate magic links for candidates.

- Exchange magic link token for a prefilled MailingListAddMember

Adds an endpoint that accepts a valid magic link token and returns a pre-filled `MailingListAddMember` model.

### DIFF
--- a/GetIntoTeachingApi/Services/CandidateMagicLinkTokenService.cs
+++ b/GetIntoTeachingApi/Services/CandidateMagicLinkTokenService.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Linq;
+using System.Security.Cryptography;
+using GetIntoTeachingApi.Models;
+
+namespace GetIntoTeachingApi.Services
+{
+    public class CandidateMagicLinkTokenService : ICandidateMagicLinkTokenService
+    {
+        private readonly ICrmService _crm;
+        private readonly RNGCryptoServiceProvider _cryptoService;
+
+        public CandidateMagicLinkTokenService(ICrmService crm)
+        {
+            _cryptoService = new RNGCryptoServiceProvider();
+            _crm = crm;
+        }
+
+        public void GenerateToken(Candidate candidate)
+        {
+            candidate.MagicLinkToken = CreateToken();
+            candidate.MagicLinkTokenCreatedAt = DateTime.UtcNow;
+        }
+
+        public Candidate Exchange(string token)
+        {
+            var matchingCandidates = _crm.MatchCandidates(token);
+
+            // Return null if there are no matches and also in the very
+            // unlikely case a token has been duplicated.
+            if (matchingCandidates.Count() != 1)
+            {
+                return null;
+            }
+
+            return matchingCandidates.First();
+        }
+
+        private string CreateToken()
+        {
+            byte[] bytes = new byte[16];
+            _cryptoService.GetBytes(bytes);
+
+            return BitConverter.ToString(bytes).Replace("-", string.Empty);
+        }
+    }
+}
--- a/GetIntoTeachingApi/Services/CrmService.cs
+++ b/GetIntoTeachingApi/Services/CrmService.cs
@@ -87,20 +87,49 @@ namespace GetIntoTeachingApi.Services
                 return null;
             }
 
-            _service.LoadProperty(entity, new Relationship("dfe_contact_dfe_candidatequalification_ContactId"), context);
-            _service.LoadProperty(entity, new Relationship("dfe_contact_dfe_candidatepastteachingposition_ContactId"), context);
-            _service.LoadProperty(entity, new Relationship("dfe_contact_dfe_servicesubscription_contact"), context);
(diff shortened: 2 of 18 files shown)


GetIntoTeachingApi/Services/CrmService.cs
```diff
@@ -87,20 +87,49 @@ public Candidate MatchCandidate(ExistingCandidateRequest request)
...
 
             return new Candidate(entity, this);
         }
 
+        public IEnumerable<Candidate> MatchCandidates(string magicLinkToken)
+        {
+            // Avoids a potentially very expensive query.
+            if (string.IsNullOrEmpty(magicLinkToken))
+            {
+                return new Candidate[0];
```
jebw, line 100: My reading of this is its an empty enumerable of Candidate models - have I got that right?

GetIntoTeachingApi/Services/CrmService.cs
```diff
@@ -87,20 +87,49 @@ public Candidate MatchCandidate(ExistingCandidateRequest request)
...
+                LoadCandidateRelationships(entity, context);
+            }
+
+            return entities.Select(e => new Candidate(e, this));
+        }
+
         public Candidate GetCandidate(Guid id)
         {
-            var entity = _service.CreateQuery("contact", Context())
-                .FirstOrDefault(c => c.GetAttributeValue<EntityReference>("contactid").Id == id);
```
jebw, line 101: will `FirstOrDefault` return `null` when there are no matches? ie is 'default' `null`

GetIntoTeachingApi/Services/CandidateMagicLinkTokenService.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            return matchingCandidates.First();
+        }
+
+        private string CreateToken()
+        {
+            byte[] bytes = new byte[16];
+            _cryptoService.GetBytes(bytes);
+
+            return BitConverter.ToString(bytes).Replace("-", string.Empty);
```
jebw, line 44: Does anything ensure this will be a url safe token?

jebw (commented):
Looks good to me - I've a couple of minor questions attached first though, thanks

GetIntoTeachingApi/Services/CrmService.cs
ethax-ross: Yeah exactly, we would never want to consider an empty token as a 'match' so returning an empty array here as 'no matches' - considered raising an exception instead but it just makes more logic to deal with that case in the same way further up the chain

GetIntoTeachingApi/Services/CrmService.cs
ethax-ross: Yep, there's `First()` which raises when there are no results and `FirstOrDefault()` which returns `null` if there are no matches

GetIntoTeachingApi/Services/CandidateMagicLinkTokenService.cs
ethax-ross: Yeah so this will always return a hex string which is URL safe:

> ToString(Byte[]) |  Converts the numeric value of each element of a specified array of bytes to its equivalent hexadecimal string representation.

jebw (approved):
LGTM

PR #485, opened 2021-02-09
Expire magic link tokens after 48hrs/first use

ethax-ross:

This PR deals with the third step of migrating existing candidates onto the new privacy policy via a mailing list sign up in the GiT app:

- [x] Add new magic link token endpoints (creating tokens and exchanging for mailing list candidate data)
- [x] Expire magic link tokens after 48 hours and ensure they are only single-use tokens. Write to field indicating when token is generated/used.
- [ ] Add CRM client and restrict access to magic link generation endpoint. 
- [ ] Add job to process pending magic link generations. Connect models to CRM attributes (need CRM changes in prod first)

---

- Set expiry date on magic link tokens

Change the `createdAt` date to be `expiresAt` and set to 48 hours in the future when creating a magic link token.

Check that the token is valid before responding with a pre-filled `MailingListAddMember` model.

If the magic link could not be matched to a candidate (or if it matches to multiple candidates) we return `BadRequest` with a status of `Invalid`.

If a token was matched to a candidate but the expiry date has passed we return `BadRequest` with a status of `Expired`.

- Prevent multiple use of magic link tokens

Update the `MagicLinkTokenStatusId` on generating and exchanging tokens.

Check that a token has not already been exchanged and return `BadRequest` with a status of `AlreadyExchanged` if it has been used already.

- Encapsulate magic link token exchange into result model

Shifts some of the logic from the controller that determines the state of a magic link token (expired/valid/already exchanged/invalid) into a result model. Change to return `Unauthorized` instead of `BadRequest` with an applicable `status` of `Invalid`, `Expired` or `AlreadyExchanged`.

### DIFF
--- a/GetIntoTeachingApi/Controllers/MailingListController.cs
+++ b/GetIntoTeachingApi/Controllers/MailingListController.cs
@@ -116,6 +116,11 @@ namespace GetIntoTeachingApi.Controllers
                 return BadRequest(new { Message = "Magic link token has expired.", Status = "Expired" });
             }
 
+            if (candidate.MagicLinkTokenAlreadyExchanged())
+            {
+                return BadRequest(new { Message = "Magic link token has already been exchanged.", Status = "AlreadyExchanged" });
+            }
+
             return Ok(new MailingListAddMember(candidate));
         }
     }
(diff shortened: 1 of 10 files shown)


jebw (approved):
LGTM

PR #489, opened 2021-02-10
Remove endpoint to create magic link tokens

ethax-ross:
We are going to generate these in a background job that looks for all candidates with a `MagicLinkTokenStatusId` of `Pending` instead of having the CRM request candidates explicitly.

Also removes unused `GetCandidates` method on `ICrmService`.


peteryates (approved):
Makes sense

PR #493, opened 2021-02-11
Database Migration

sfawcett123:
[Database Migration](https://trello.com/c/7DrTCJwi/1128-invoke-high-availability-for-our-postgres-dbs)

### Objective
The current databases are not high availability, to make them HA we need to change their service plan. but we do not need HA databases on all environments.

### Changes
Added some new variables so we can have different plans for different environments defaulting to the small-11 plan
While testing created a sandbox test, thought it wise to keep the files for this in the future.

### Testing

**Test 1**
1.  Created a sandpit environment with the API and its databases.
2. In one window ran Apache Benchmark in a loop against the API Health Check endpoint
3. in another window monitored the output logs from the API
4. in a third window carried out a manual `cf update-service` command to update the database

This test proved the service was always available,

**Test 2**
1. After manual migration ( step 4 above ) ran terraform with new plan
2 Terraform did not identify any change needed

This test proved terraform did not get confused with plan change.

### Review
This PR is not the final change, a subsequent PR will need to be released POST_MIGRATION that will set the PLAN
This PR does not change the  API


ethax-ross (dismissed):
The bits I understand look fine but you may want a second pair of eyes on it

PR #495, opened 2021-02-11
Minor bug fixes for magic link mechanism

ethax-ross:
I missed these originally as there was no test data in the CRM available, but we have it now so its highlighted theses issues.

- Share context when loading related entities

The candidate was being loaded in a different context to the related entities, which was causing an error (you can't relate entities across contexts).

- Persist candidate after exchanging magic link token

When the magic link token gets exchanged the candidate status is updated. We weren't persisting this back to the CRM, which meant tokens could be re-used until they expired (instead of being single use).


peteryates (approved):
Looks good

PR #503, opened 2021-02-19
Sanitize invalid CRM postcodes

ethax-ross:
- Extract postcode validation into property validator

Initially I did this so that I could re-use it in the request models, but FluentValidation doesn't let you validate a single value (only an entire model).

It's a worthwhile refactoring anyway, so leaving it in.

- Sanitize invalid postcodes coming back from the CRM

We have a potential scenario where an candidate in the CRM has an invalid postcode against them; if we use their data when pre-filling one of the forms in the front-end apps this postcode will then be re-submitted to the API and it will fail validation (the data is only validated on the client-side if shown to the user and postcodes are often optional/skipped steps).

This commit updates the API to sanitize invalid postcodes coming back from the CRM.

### DIFF
--- a/GetIntoTeachingApi/Models/Candidate.cs
+++ b/GetIntoTeachingApi/Models/Candidate.cs
@@ -256,6 +256,11 @@ namespace GetIntoTeachingApi.Models
         public Candidate(Entity entity, ICrmService crm)
             : base(entity, crm)
         {
+            // Treat invalid postcodes coming back from the CRM as null.
+            if (AddressPostcode != null && !Location.PostcodeRegex.IsMatch(AddressPostcode))
+            {
+                AddressPostcode = null;
+            }
         }
 
         public bool HasTeacherTrainingAdviser()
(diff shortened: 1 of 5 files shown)


GetIntoTeachingApi/Models/Candidate.cs
```diff
@@ -256,6 +256,11 @@ public Candidate()
         public Candidate(Entity entity, ICrmService crm)
             : base(entity, crm)
         {
+            // Treat invalid postcodes coming back from the CRM as null.
+            if (AddressPostcode != null && !Location.PostcodeRegex.IsMatch(AddressPostcode))
```
jebw, line 260: This looks like it duplicates the logic of the Postcode validator - is it possible to use that here instead to DRY it out?

jebw (approved):

GetIntoTeachingApi/Models/Candidate.cs
ethax-ross: Sadly not - I thought the same thing which is why I initially extracted the postcode validation into its own property validator class, but it appears FluentValidation isn't geared towards validating individual values, only complete models. To use it against a value would mean constructing the validation context as I have in the tests, which is pretty convoluted:

https://github.com/DFE-Digital/get-into-teaching-api/pull/503/files#diff-283211bed6243fefde5cf64dfaa7649210c5b7219a7c48286204d2d364aff410R32-R36

The only other option that may work is injecting the `CandidateValidator` into the model and asking that to validate only the postcode attribute, but given the constructor is called via reflection I'd rather not complicate the logic further by injecting the appropriate validator into the model when its creating them. If this becomes a pattern we need to deal with in general though (invalid data in the CRM that needs to be sanitized) I'll take a look at doing it in a DRY-er way across all models then 👍

PR #511, opened 2021-02-24
Support explicitly setting the type on the sign up request model

ethax-ross:
Currently the `TypeId` of a `Candidate` is inferred based on if any past teaching positions are present. Instead, we can explicitly pass this from the client in response to the "Are you returning to teaching?" question. This will have two benefits:

- We can then pre-fill this question on matchback
- We can use `TypeId` going forward to determine if a candidate is  returning to   teaching

Initially this is being added as an optional field; once the TTA service has been updated we can add validation to ensure it is always present and remove the code to infer the type.


peteryates (approved):
Looks good.

PR #513, opened 2021-02-24
Stop inferring returning teacher status from other attributes

ethax-ross:
Until recently determining if a candidate was returning to teaching (vs interested in teaching) involved checking the `Candidate` model for `PastTeachingPosition` or checking the `TeacherTrainingAdviserSignUp` model for a `SubjectTaughtId`. This isn't very intuitive and the CRM actually stores a `TypeId` that tells us if the candidate is a returner. Now that the TTA service explicitly sends the `TypeId` we can consolidate and simplify how we determine if a `Candidate` is a returner.

- Make TypeId required on TeacherTrainingAdviserSignUp

The TTA app has been updated to send a `type_id` with all requests now, so we can mark this as required and remove the inferred logic around candidate type.

- Stop inferring candidate type in TeacherTrainingAdviserSignUp

This is now passed explicitly from the client and marked as required in the TeacherTrainingAdviserSignUp, so we no longer need to infer it.

Also updates `Candidate` to use the `TypeId` instead of inferring returning teachers from if they have any `PastTeachingPositions`.

Various tests updated so that they make more sense (instead of setting `SubjectTaughtId` or `PastTeachingPositions` to indicate a returner the `TypeId` is now explicitly set).

### DIFF
--- a/GetIntoTeachingApi/Models/Candidate.cs
+++ b/GetIntoTeachingApi/Models/Candidate.cs
@@ -270,7 +270,7 @@ namespace GetIntoTeachingApi.Models
 
         public bool IsReturningToTeaching()
         {
-            return PastTeachingPositions.Count > 0;
+            return TypeId == (int)Type.ReturningToTeacherTraining;
         }
 
         public bool HasGcseMathsAndEnglish()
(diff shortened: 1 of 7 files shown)


peteryates (approved):
Seems sensible.

PR #517, opened 2021-03-01
Add change tracking to models/only update CRM with changed attributes

ethax-ross:
The API currently skips writing `null` values back to the CRM (for reasons detailed below). Instead, we want to add change tracking support to the models so that we can instead write back any value that has changed and ignore unchanged attributes. There's a more detailed description below, but the best way to implement this I found was using [Fody](https://github.com/Fody/Fody) with the [PropertyChanged weaver](https://github.com/Fody/PropertyChanged) to dynamically adhere to `INotifyPropertyChanged ` at compile-time.

- Move mock models into main project

I'm not sure why, but even though the `MockModel` inherits from `BaseModel` Fody is not modifying it at compile-time to weave in the property change updates.

Moving the mocks into the main project fixes this for now and I'll investigate it further in a follow up PR.

- Track property changes on models

Implementing dirty tracking of model attributes in C# is hard. There are a couple of ways you can do this; initially I tried using reflection to dynamically override setters and track their usage. Whilst this is possible, it requires making all properties `virtual` and injecting a proxy class that handles hijacking the setter methods. This had the side-effect of making a bunch of tests fail (as FluentValidation validators are tied to the model types) and also felt fickle; if a property was added and not virtual it would be ignored.

Another approach is to use the `INotifyPropertyChanged` interface which is typically used in C# MVVM patterns to drive view updates. This requires manually overriding each accessor and calling the change notifier; again, not ideal and prone to errors in the future if someone forgets to do this for their new accessor.

This commit uses `INotifyPropertyChanged` in conjunction with Fody to allow the setters to be modified at compile-time using the `Property` weaver. It means as long as our class implements `INotifyPropertyChanged` it will have the relevant
accessor code injected in at compile time to ensure change are picked up.

The `BaseModel` has been updated to include a `HashSet` of `ChangedPropertyNames`, which we can use later on in order to ensure only changed attributes are written to the CRM.

Hangfire automatically serializes payload objects using `Newtonsoft.Json`, which doesn't deserialize the `ChangedPropertyNames` accurately (it modifies the value as its deserializing, so you end up with all attributes being marked
as changed). To fix this the deserialization lifecycle is hooked into and change tracking is disabled during that process. `System.Text.Json` appears to deserialize the payloads fine, but I've included tests to ensure we know if that changes in the future.

- Only write changed attributes back to the CRM

Currently, the API will write all model attributes back to the CRM, skipping `null` values entirely. This approach was taken due to having multiple request models mapping back to the same `Candidate` model that gets persisted to the CRM -- each request model contains a subset of `Candidate` attributes and we don't want to `null`-out the attributes that weren't present in the request model.

We had no way of distinguishing between a `Candidate` attribute that has been set to `null` from a request model and a `Candidate` attribute that has been unchanged by the request model.

To address this the base models have been updated to include change tracking information. Now when we map from a request model to, say, a `Candidate`, we maintain a `HashSet` of the changed property names. The mapping code now
references these and skips any attributes that have not been changed (also writing any that have been changed and are now `null`).

When the `Candidate` is mapped to a CRM `Entity` the entity will maintain its own change tracking and ensures only changed properties are written back to the CRM.

- Don't change channel for existing candidates

When an existing candidate has matched-back and gone through one of our 3 transactions we do not want to change their `ChannelId` in the CRM. Previously we would set this to `null` and `null` values were not written to the CRM, but now we write `null` values we need to instead not change the `ChannelId` (we only write changed attributes to the CRM now).

Includes minor change to ensure a candidate qualification is not created on event sign up unless they opt-in to the mailing list and specify their degree subject.

Removes redundant setting of education phase for returner when setting past teaching positions; this is already set in another method of this class.


GetIntoTeachingApi/Models/BaseModel.cs
```diff
@@ -27,10 +38,13 @@ public BaseModel(Entity entity, ICrmService crm, IValidatorFactory vaidatorFacto
 
             MapFieldAttributesFromEntity(entity);
             MapRelationshipAttributesFromEntity(entity, crm, vaidatorFactory);
-
             NullifyInvalidFieldAttributes(vaidatorFactory);
-        }
-
+        }
+
+        #pragma warning disable 67
```
peteryates, line 44: I didn't even realise you could do this in C# 🤯 

Vague memories of seeing it in C++ some decades ago.

peteryates (approved):
Overall looks good to me and the approach seems sensible.

PR #519, opened 2021-03-02
Record metrics on event search results

ethax-ross:
[Trello-1352](https://trello.com/c/wAptzepa/1352-spike-review-the-radius-of-the-events-finder-to-10-miles-30-miles-50-miles)

We are thinking about tweaking the event search radius select options in the GiT app from the current options of 30/50/100 miles to see if users engage better with different search radii. Before we change it, we want to understand how users are interacting with the search in terms of which radius they select and how many results they receive in their search.

To enable this, we are adding a new histogram metric that observes event search results with labels for type and radius; this should give insight into how the different types of searches are performing.


peteryates (approved):
Nice.

PR #521, opened 2021-03-03
Validate Azure Keys

sfawcett123:
### [Validate Keys ](https://trello.com/c/YTWSmEQL/1341-check-secrets-before-deploy)

Validate the azure keys before usage, to prevent badly formed keys from causing issues further down the pipeline.

### Review
This process does not effect the API, it is a step that prevents deployment if the key was invalid. The deployment probably would have failed anyway due to the invalid key, but under certain circumstances the error with the key details could get logged.

### Tested
https://github.com/DFE-Digital/get-into-teaching-api/runs/2021154429?check_suite_focus=true

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -11,10 +11,32 @@ on:
 
 
 jobs:
+  validate:
+    name: Validate Secrets
+    runs-on: ubuntu-latest
+    environment: 
+       name: ${{ github.event.inputs.environment }}
+    steps:
+       - name: Checkout
+         uses: actions/checkout@v2
+
+       - uses: azure/login@v1
+         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -11,10 +11,32 @@ on:
...
 
 jobs:
+  validate:
+    name: Validate Secrets
+    runs-on: ubuntu-latest
+    environment: 
+       name: ${{ github.event.inputs.environment }}
+    steps:
+       - name: Checkout
+         uses: actions/checkout@v2
```
vigneshmsft, line 21: don't think checkout is required

vigneshmsft (approved):

PR #522, opened 2021-03-03
Manually serialize Hangfire payloads

ethax-ross:
Hangfire uses Newtonsoft.Json by default and it turns out this wasn't deserializing payloads with change tracking information correctly - this PR swaps it out for System.Text.Json and makes some updates to ensure this is deserializing consistently.

Also marks properties defined with a value as changed on initialisation, as we do this for default values in models and they were being missed/skipped over.

- Remove custom Newtonsoft.Json deserialization logic

We're going to manually serialize/deserialize the Hangfire payloads using the default System.Text.Json serializer used in the rest of the app, instead of relying on Newtonsoft.Json that Hangfire uses by default (it appears to be buggy around deserializing the `ChangedPropertyNames` attribute of nested models).

- Mark properties defined with a value as changed

In a number of places we define attributes with a value, for example:

```
public string FieldWithValue { get; set; } = "initial value";
```

As these are set as part of initialising the object we don't receive property changed notifications. Instead, we have to manually fire the notifications on initialisation for these fields.

- Fix bug in deserializing ChangedPropertyNames

Initially I thought System.Text.Json was correctly deserializing `ChangedPropertyNames` by default, but it turns out when you get more complex objects with nesting and inheritance it can break due to the way the serializer works. It was also dependent on the order of the attributes when serialized.

In order to consistently serialize/deserialize models with change tracking this commit:

\- Turns on `IgnoreNullValues` - if the serializer encounters a `null` we don't want it to deserialize the attribute and modify `ChangedPropertyNames`. If a `null` _was_ changed it will already be present in the list when deserialized anyway. This will also keep the Hangifre payload length down and make it easier to interpret.

\- Turn change tracking off during deserialization. We serialize  `ChangedPropertyNames` so we don't want the deserializer to inadvertently effect the values.

\- Improves test coverage around deserialization to ensure various   orders/combinations work correctly.

- Manually serialize/deserialize UpsertCandidateJob payload

If we pass an object to a Hangfire job the library serializes/deserilizes it for us using Newtonsoft.Json. This library is buggy when it comes to deserializing the change tracking information and we also use System.Text.Json everywhere else in the application, so it makes sense to do it manually/using the same library. It also means we can have test coverage for the
serialization/deserialization process.

- Add convenience extensions for serializing/deserializing change tracked models

Add an extension that loads in the necessary `JsonSerializerOptions` to correctly serialize/deserialize models that have change tracking enabled.

### DIFF
--- a/GetIntoTeachingApi/Controllers/MailingListController.cs
+++ b/GetIntoTeachingApi/Controllers/MailingListController.cs
@@ -1,10 +1,10 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Text.Json;
 using GetIntoTeachingApi.Attributes;
 using GetIntoTeachingApi.Jobs;
 using GetIntoTeachingApi.Models;
 using GetIntoTeachingApi.Services;
+using GetIntoTeachingApi.Utils;
 using Hangfire;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
@@ -60,7 +60,7 @@ namespace GetIntoTeachingApi.Controllers
                 return BadRequest(this.ModelState);
             }
 
-            string json = JsonSerializer.Serialize(request.Candidate);
+            string json = request.Candidate.SerializeChangedTracked();
             _jobClient.Enqueue<UpsertCandidateJob>((x) => x.Run(json, null));
 
             return NoContent();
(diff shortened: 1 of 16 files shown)


peteryates (approved):
All seems sensible to me.

PR #524, opened 2021-03-04
Swap Newtonsoft.Json to System.Text.Json

ethax-ross:
We still use Newtonsoft.Json in a couple of places; ideally we want to use System.Text.Json everywhere so we only have the one Json serializer; this updates the remaining places.

Updates integers in JSON fixtures to be without quotes as System.Text.Json is stricter than Newtonsoft.Json (the VCAP_SERVICES has them without quotes in cloud foundry so this change shouldn't effect the environments).


peteryates (approved):
Looks good

PR #527, opened 2021-03-04
Teaching event buildings

jkempster34:
A new form will be added to the app to allow Abramis to add teaching events, part of this form needs to produce a list of teaching event buildings. Currently, the API only retrieves buildings that are associated with an event.

This PR:
- Adds a new endpoint to get all buildings from the CMS.
- Updates the CMS/postgres sync to:
  - get events without their associated buildings from the CMS.
  - get all buildings before associating them with events via BuildingId.

### DIFF
--- a/GetIntoTeachingApi/Controllers/TeachingEventBuildingsController.cs
+++ b/GetIntoTeachingApi/Controllers/TeachingEventBuildingsController.cs
@@ -9,7 +9,6 @@ namespace GetIntoTeachingApi.Controllers
 {
     [Route("api/teaching_event_buildings")]
     [ApiController]
-    [LogRequests]
     [Authorize(Roles = "Admin,GetIntoTeaching")]
     public class TeachingEventBuildingsController : ControllerBase
     {
--- a/GetIntoTeachingApi/Controllers/TeachingEventBuildingsController.cs
+++ b/GetIntoTeachingApi/Controllers/TeachingEventBuildingsController.cs
@@ -1,0 +1,39 @@
+﻿using GetIntoTeachingApi.Attributes;
+using GetIntoTeachingApi.Models;
+using GetIntoTeachingApi.Services;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Swashbuckle.AspNetCore.Annotations;
+
+namespace GetIntoTeachingApi.Controllers
+{
+    [Route("api/teaching_event_buildings")]
+    [ApiController]
+    [LogRequests]
+    [Authorize(Roles = "Admin,GetIntoTeaching")]
+    public class TeachingEventBuildingsController : ControllerBase
+    {
+        private readonly IStore _store;
+
+        public TeachingEventBuildingsController(IStore store)
+        {
+            _store = store;
+        }
+
+        [HttpGet]
+        [CrmETag]
+        [PrivateShortTermResponseCache]
+        [Route("teaching_event_buildings")]
+        [SwaggerOperation(
+            Summary = "Retrieves all event buildings.",
+            OperationId = "GetTeachingEventBuildings",
+            Tags = new[] { "Teaching Events" })]
+        [ProducesResponseType(typeof(TeachingEventBuilding), 200)]
+        public IActionResult GetTeachingEventBuildings()
+        {
+            var buildings = _store.GetTeachingEventBuildings();
+
+            return Ok(buildings);
+        }
+    }
+}
--- a/GetIntoTeachingApi/Controllers/TeachingEventBuildingsController.cs
+++ b/GetIntoTeachingApi/Controllers/TeachingEventBuildingsController.cs
@@ -10,6 +10,7 @@ namespace GetIntoTeachingApi.Controllers
     [Route("api/teaching_event_buildings")]
     [ApiController]
     [Authorize(Roles = "Admin,GetIntoTeaching")]
+    [PrivateShortTermResponseCache]
     public class TeachingEventBuildingsController : ControllerBase
     {
         private readonly IStore _store;
@@ -21,7 +22,6 @@ namespace GetIntoTeachingApi.Controllers
 
         [HttpGet]
         [CrmETag]
-        [PrivateShortTermResponseCache]
         [Route("")]
         [SwaggerOperation(
             Summary = "Retrieves all event buildings.",
--- a/GetIntoTeachingApi/Controllers/TeachingEventBuildingsController.cs
+++ b/GetIntoTeachingApi/Controllers/TeachingEventBuildingsController.cs
@@ -22,11 +22,11 @@ namespace GetIntoTeachingApi.Controllers
         [HttpGet]
         [CrmETag]
         [PrivateShortTermResponseCache]
-        [Route("teaching_event_buildings")]
+        [Route("")]
         [SwaggerOperation(
             Summary = "Retrieves all event buildings.",
             OperationId = "GetTeachingEventBuildings",
-            Tags = new[] { "Teaching Events" })]
+            Tags = new[] { "Teaching Event Buildings" })]
         [ProducesResponseType(typeof(TeachingEventBuilding), 200)]
         public IActionResult GetTeachingEventBuildings()
         {
--- a/GetIntoTeachingApi/Controllers/TeachingEventsController.cs
+++ b/GetIntoTeachingApi/Controllers/TeachingEventsController.cs
@@ -97,22 +97,6 @@ namespace GetIntoTeachingApi.Controllers
             return Ok(teachingEvent);
         }
 
-        [HttpGet]
-        [CrmETag]
-        [PrivateShortTermResponseCache]
-        [Route("teaching_event_buildings")]
-        [SwaggerOperation(
-         Summary = "Retrieves all event buildings.",
-         OperationId = "GetTeachingEvent",
-         Tags = new[] { "Teaching Events" })]
-        [ProducesResponseType(typeof(TeachingEventBuilding), 200)]
-        public IActionResult GetTeachingEventBuildings()
-        {
-            var teachingEvents = _store.GetTeachingEventBuildings();
-
-            return Ok(teachingEvents);
-        }
-
         [HttpPost]
         [Route("attendees")]
         [SwaggerOperation(
--- a/GetIntoTeachingApi/Controllers/TeachingEventsController.cs
+++ b/GetIntoTeachingApi/Controllers/TeachingEventsController.cs
@@ -97,6 +97,22 @@ namespace GetIntoTeachingApi.Controllers
             return Ok(teachingEvent);
         }
 
+        [HttpGet]
+        [CrmETag]
+        [PrivateShortTermResponseCache]
+        [Route("teaching_event_buildings")]
+        [SwaggerOperation(
+         Summary = "Retrieves all event buildings.",
+         OperationId = "GetTeachingEvent",
+         Tags = new[] { "Teaching Events" })]
+        [ProducesResponseType(typeof(TeachingEventBuilding), 200)]
+        public IActionResult GetTeachingEventBuildings()
+        {
+            var teachingEvents = _store.GetTeachingEventBuildings();
+
+            return Ok(teachingEvents);
+        }
+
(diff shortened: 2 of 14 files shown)


GetIntoTeachingApi/Controllers/TeachingEventsController.cs
```diff
@@ -91,6 +91,22 @@ public async Task<IActionResult> Get([FromRoute, SwaggerParameter("The `readable
...
+        [HttpGet]
+        [CrmETag]
+        [PrivateShortTermResponseCache]
+        [Route("teaching_event_buildings")]
+        [SwaggerOperation(
+         Summary = "Retrieves all event buildings.",
+         OperationId = "GetTeachingEvent",
+         Tags = new[] { "Teaching Events" })]
+        [ProducesResponseType(typeof(TeachingEventBuilding), 200)]
+        public IActionResult GetTeachingEventBuildings()
```
ethax-ross, line 103: As we're going to have a couple of endpoints here (there will be another for creating a building) I would be inclined to put this in its own `TeachingEventBuildingsController`. Also typo in the `OperationId` should be `GetTeachingEventBuildings`.

GetIntoTeachingApi/Controllers/TeachingEventsController.cs
```diff
@@ -91,6 +91,22 @@ public async Task<IActionResult> Get([FromRoute, SwaggerParameter("The `readable
...
+        {
+            var teachingEvents = _store.GetTeachingEventBuildings();
```
ethax-ross, line 105: Type in the variable name, `teachingEventBuildings`

GetIntoTeachingApi/Models/TeachingEvent.cs
```diff
@@ -68,6 +68,7 @@ public enum EventType
         [EntityRelationship("msevtmgt_event_building", typeof(TeachingEventBuilding))]
         public TeachingEventBuilding Building { get; set; }
         [JsonIgnore]
+        [EntityField("msevtmgt_building", typeof(EntityReference), "msevtmgt_buildingid")]
```
ethax-ross, line 71: Can we add a test for this attribute? there should be others you can see in `TeachingEventTests` that are tested to copy

GetIntoTeachingApi/Services/CrmService.cs
```diff
@@ -222,14 +222,19 @@ public IEnumerable<TeachingEvent> GetTeachingEvents(DateTime? startAfter = null)
             query.Criteria.AddFilter(filter);
 
             var link = query.AddLink("msevtmgt_building", "msevtmgt_building", "msevtmgt_buildingid", JoinOperator.LeftOuter);
```
ethax-ross, line 224: The entire `link` should be dropped here as well, you've just removed the columns setting for the link with this change. Did you check if GetTeachingEvents is used elsewhere? I don't think it is, but if it is we may need to include buildings for other scenarios and exclude for the sync scenario.

GetIntoTeachingApi/Services/Store.cs
```diff
@@ -145,21 +151,22 @@ private async Task<IEnumerable<TeachingEvent>> OnlineEventsMatchingRequest(Teach
...
-            var buildings = teachingEvents.Where(te => te.Building != null)
-                .Select(te => te.Building).DistinctBy(b => b.Id);
-
-            await SyncModels(buildings, _dbContext.TeachingEventBuildings);
 
-            // Link events with buildings attached to the context prior to sync.
             foreach (var te in teachingEvents.Where(te => te.Building != null))
             {
-                te.Building = await _dbContext.TeachingEventBuildings.FindAsync(te.Building.Id);
+                te.Building = await _dbContext.TeachingEventBuildings.FindAsync(te.BuildingId);
```
ethax-ross, line 169: It may be more efficient to load all buildings into memory and run `Find` on that instead of hitting the database for each teaching event

GetIntoTeachingApiTests/Services/StoreTests.cs
```diff
@@ -472,11 +491,24 @@ public async void GetTeachingEventAsync_WithId_ReturnsMatchingEvent()
...
             result.ReadableId.Should().Be(events.First().ReadableId);
             result.Building.Should().NotBeNull();
+        }
+
+        [Fact]
+        public async Task GetTeachingEventBuildings_ReturnsAll()
+        {
+            await SeedMockTeachingEventBuildingsAsync();
+            var events = await SeedMockTeachingEventsAsync();
+            _mockCrm.Setup(m => m.GetTeachingEvents(It.IsAny<DateTime>())).Returns(events);
```
ethax-ross, line 507: Not following this test, why are we mocking `GetTeachingEvents` when we're querying the buildings?

GetIntoTeachingApiTests/Services/StoreTests.cs
```diff
@@ -500,10 +533,7 @@ private static IEnumerable<TeachingEvent> MockTeachingEvents()
                 TypeId = (int)TeachingEvent.EventType.TrainToTeachEvent,
                 IsOnline = true,
                 StartAt = DateTime.UtcNow.AddDays(5),
-                Building = new TeachingEventBuilding()
-                {
-                    Id = sharedBuildingId,
-                }
+                BuildingId = sharedBuildingId
```
ethax-ross, line 536: Instead of having all the `BuildingId`s as constants can we just have something like:

```
var buildings = MockTeachingEventBuildings();
var sharedBuildingId = buildings[0];

...
BuildingId = buildings[1]
...
BuildingId = buildings[2]
```

GetIntoTeachingApiTests/Services/StoreTests.cs
```diff
@@ -96,26 +104,25 @@ public async void SyncAsync_UpdatesExistingTeachingEvents()
         }
 
         [Fact]
-        public async void SyncAsync_DeletesOrphanedTeachingEventsAndBuildings()
+        public async void SyncAsync_DeletesOrphanedTeachingEvents()
         {
+            await SeedMockTeachingEventBuildingsAsync();
```
ethax-ross, line 109: Instead of doing this everywhere:

```
await SeedMockTeachingEventBuildingsAsync();
await SeedMockTeachingEventsAsync();
```

Can we just change `SeedMockTeachingEventsAsync` to `SeedMockTeachingEventsAndBuildingsAsync` and call the buildings seed from inside there to DRY it up a bit?

GetIntoTeachingApiTests/Services/StoreTests.cs
```diff
@@ -25,6 +25,12 @@ public class StoreTests : DatabaseTests
         private readonly Mock<IGeocodeClientAdapter> _mockGeocodeClient;
```
ethax-ross, line 25: I think we need to split the teaching event building sync specs out of the teaching event sync specs now as the behaviour has changed (and we sync buildings without events whereas the existing tests only cover buildings with events), so have:

```
public async void SyncAsync_InsertsNewTeachingEventBuildings()
public async void SyncAsync_UpdatesExistingTeachingEvents()
public async void SyncAsync_DeletesOrphanedTeachingEventBuildings()
```

ethax-ross (commented):
Looking good, few suggestions left

GetIntoTeachingApi/Controllers/TeachingEventBuildingsController.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public TeachingEventBuildingsController(IStore store)
+        {
+            _store = store;
+        }
+
+        [HttpGet]
+        [CrmETag]
+        [PrivateShortTermResponseCache]
+        [Route("teaching_event_buildings")]
```
ethax-ross, line 25: This should be:

```
[Route("")]
```

Otherwise your path will be `/api/teaching_event_building/teaching_event_buildings`

GetIntoTeachingApi/Controllers/TeachingEventBuildingsController.cs
```diff
@@ -0,0 +1,38 @@
...
+        [SwaggerOperation(
+            Summary = "Retrieves all event buildings.",
+            OperationId = "GetTeachingEventBuildings",
+            Tags = new[] { "Teaching Events" })]
```
ethax-ross, line 29: Teaching Event Buildings

GetIntoTeachingApi/Services/Store.cs
```diff
@@ -145,21 +151,23 @@ private async Task<IEnumerable<TeachingEvent>> OnlineEventsMatchingRequest(Teach
...
         {
             var afterDate = DateTime.UtcNow.Subtract(TeachingEventArchiveSize);
             var teachingEvents = _crm.GetTeachingEvents(afterDate).ToList();
-            await PopulateTeachingEventCoordinates(teachingEvents);
-
-            var buildings = teachingEvents.Where(te => te.Building != null)
-                .Select(te => te.Building).DistinctBy(b => b.Id);
-
-            await SyncModels(buildings, _dbContext.TeachingEventBuildings);
+            var teachingEventBuildings = _crm.GetTeachingEventBuildings();
```
ethax-ross, line 166: This should load the buildings from the store as we've already cached them in the previous sync operation

GetIntoTeachingApiTests/Services/StoreTests.cs
```diff
@@ -96,26 +96,54 @@ public async void SyncAsync_UpdatesExistingTeachingEvents()
...
         {
-            await SeedMockTeachingEventsAsync();
+            await SeedMockTeachingEventsAndBuildingsAsync();
             var teachingEvents = MockTeachingEvents().ToList();
             _mockCrm.Setup(m => m.GetTeachingEvents(It.Is<DateTime>(d => CheckGetTeachingEventsAfterDate(d)))).Returns(teachingEvents.GetRange(0, 1));
 
             await _store.SyncAsync();
 
             DbContext.TeachingEvents.Should().BeEquivalentTo(teachingEvents.GetRange(0, 1));
-            DbContext.TeachingEventBuildings.Should()
```
ethax-ross, line 108: Sorry I maybe said to remove this in a previous comment (can't remember) but it's actually testing that when teaching events are saved they get correctly connected to the buildings (its been flagged as missing coverage by SonarCloud)

GetIntoTeachingApiTests/Services/StoreTests.cs
```diff
@@ -472,11 +501,22 @@ public async void GetTeachingEventAsync_WithId_ReturnsMatchingEvent()
...
 
             result.ReadableId.Should().Be(events.First().ReadableId);
             result.Building.Should().NotBeNull();
+        }
+
+        [Fact]
+        public async Task GetTeachingEventBuildings_ReturnsAll()
+        {
+            var buildings = await SeedMockTeachingEventBuildingsAsync();
+            _mockCrm.Setup(m => m.GetTeachingEventBuildings()).Returns(buildings);
```
ethax-ross, line 515: We shouldn't need to mock the CRM here as the method loads buildings from the store

GetIntoTeachingApiTests/Controllers/TeachingEventBuildingsControllerTests.cs
```diff
@@ -0,0 +1,41 @@
...
+using GetIntoTeachingApi.Services;
+using Microsoft.AspNetCore.Mvc;
+using Moq;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace GetIntoTeachingApiTests.Controllers
+{
+    public class TeachingEventBuildingsControllerTests
```
ethax-ross, line 13: Can we add a couple more tests here for the `Authorize` attribute and the caching attributes?

https://github.com/DFE-Digital/get-into-teaching-api/blob/master/GetIntoTeachingApiTests/Controllers/TeachingEventsControllerTests.cs#L49-L71

^ there's actually a typo in that last spec it should read `PrivateShortTermResponseCache_IsPresent()`

ethax-ross (commented):
Looks good - there's a couple of attribute tests that could be added to the controller and I think we're still missing a test for `SyncAsync_DeletesOrphanedTeachingEventBuildings` to ensure if a building is removed in the CRM we remove it from our cache?

ethax-ross (approved):
LGTM

PR #528, opened 2021-03-04
Format postcodes before sending to the API

ethax-ross:
- Rename extension for clarity

- Add StringExtensions for formatting postcodes

The extension will take a valid postcode and ensure it is formatted correctly (upper case and with a space before the inward part of the postcode).

- Format postcodes from request models

When a postcode comes into the API we want to format it before it gets to the CRM; this adds formatting when mapping from the request models to a `Candidate`.

### DIFF
--- a/GetIntoTeachingApiTests/Utils/StringExtensionsTests.cs
+++ b/GetIntoTeachingApiTests/Utils/StringExtensionsTests.cs
@@ -10,6 +10,7 @@ namespace GetIntoTeachingApiTests.Utils
         [InlineData("KY11 9YU", "KY11 9YU")]
         [InlineData("KY119YU", "KY11 9YU")]
         [InlineData("ca48le", "CA4 8LE")]
+        [InlineData("M12WD", "M1 2WD")]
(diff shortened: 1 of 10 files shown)


GetIntoTeachingApiTests/Utils/StringExtensionsTests.cs
```diff
@@ -0,0 +1,22 @@
...
+    public class StringExtensionsTests
+    {
+        [Theory]
+        [InlineData("KY11 9YU", "KY11 9YU")]
+        [InlineData("KY119YU", "KY11 9YU")]
+        [InlineData("ca48le", "CA4 8LE")]
+        [InlineData("ca4 8LE", "CA4 8LE")]
+        [InlineData("  ca4   8L E", "CA4 8LE")]
+        [InlineData("invalid", null)]
+        [InlineData(null, null)]
```
peteryates, line 16: Worth adding a test for postcodes with a single district code, eg `M1 2WD`?

peteryates (dismissed):
Looks good, left a minor suggestion that might be worth adding.

peteryates (approved):

PR #530, opened 2021-03-04
Update action return types to use AspNetCore.Http.StatusCodes

jkempster34:
Use AspNetCore.Http.StatusCodes for action annotations to clean up/provide some context to the return types


ethax-ross (approved):
💯

PR #532, opened 2021-03-05
Optimise metrics by removing high-cardinality labels

ethax-ross:
- Remove high cardinality metric labels

It turns out that using high cardinality labels against metrics is bad as each combination is recorded as its own time-series, which is expensive (lots of different label values == lots of different time series).

We do this in a couple of metrics and for the most part the labels aren't actually used when it comes to graphing or alerting on the metrics, so this commit removes them.

- Remove panel using high-cardinality label

The Google requests by postcode panel is operating on a metric with a high cardinality label (postcode, which has a lot of values). Removing it as this impacts performance; the data is still sent to the logs.


peteryates (approved):
Looks good. Every time I see `totp` I think of Top of the Pops though

PR #538, opened 2021-03-08
Record in-person event search result metrics

ethax-ross:
- Collect metric on in person event result counts

When a teaching event search is performed we want to know how many in-person events are returned (as well as total events, including online).

Add a second metric that observes how many in-person events are returned in a search result.

Add convenience method to `TeachingEvent` to determine if it is `InPerson`.

- Update event search dashboard to show in-person results

We are interested in how many in-person event results are returned on radius searches (as online events are always returned, skewing the data we're interested in).

### DIFF
--- a/GetIntoTeachingApi/Models/TeachingEvent.cs
+++ b/GetIntoTeachingApi/Models/TeachingEvent.cs
@@ -70,6 +70,11 @@ namespace GetIntoTeachingApi.Models
         public Guid? BuildingId { get; set; }
         public bool IsVirtual => IsOnline && !string.IsNullOrWhiteSpace(Building?.AddressPostcode);
 
+        // The department refers to 'virtual' events as "in-person" (as
+        // well as offline events), so whilst virtual events are in fact online,
+        // they are deemed in-person here for consistency.
+        public bool IsInPerson => !IsOnline || IsVirtual;
+
         public TeachingEvent()
             : base()
         {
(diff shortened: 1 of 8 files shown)


GetIntoTeachingApi/Models/TeachingEvent.cs
```diff
@@ -70,6 +70,11 @@ public enum EventType
         public Guid? BuildingId { get; set; }
         public bool IsVirtual => IsOnline && !string.IsNullOrWhiteSpace(Building?.AddressPostcode);
 
+        // The department refers to 'virtual' events as "in-person" (as
+        // well as offline events), so whilst virtual events are in fact online,
+        // they are deemed in-person here for consistency.
+        public bool IsInPerson => !IsOnline || IsVirtual;
```
peteryates, line 76: Useful comment here, couldn't have worked out what was going on without it 😓

peteryates (approved):
Looks good to me.

GetIntoTeachingApi/Models/TeachingEvent.cs
ethax-ross: Yeah it's an odd one, I was going to go with `IsIntendedToBeInPerson` which fits the virtual events but not so much the offline events. I figure keeping it consistent with the language in use just now is probably the safest bet.

PR #540, opened 2021-03-09
Post teaching event

jkempster34:
A new form will be added to the app to add teaching events, a new endpoint needs to be added to the API to allow this.

This PR:
- Add new validator for teaching event.
- Add new endpoint to create an event.


ethax-ross (commented):
Looks good, just a couple of minor points

PR #543, opened 2021-03-12
Improved request/response and job payload logging

ethax-ross:
- Add utilty class to redact sensitive data from JSON

We want to be able to log out the request/response bodies and Hangfire job paylaods to aid in debugging when something goes wrong. While doing this we need to ensure no personally identifiable information is sent to our logstash.

Add a `Redactor` utility class that redacts sensitive data (identified by payload keys) from JSON.

Update README to explain why we have to use Newtonsoft.Json for this.

- Log out request/response payloads

Add middleware that intercepts request/response payloads, redacts sensitive data and them logs the request information.

So as not to upset other middleware in the chain the request/response body streams are copied and rewound. `RecyclableMemoryStreamManager` is used for performance/to avoid memory leaks.

Query parameters are assumed to be safe/not sensitive.

- Log out JSON job payloads

To better understand issues in production its helpful to have the full job payloads; this logs them out after redacting sensitive information.

### DIFF
--- a/GetIntoTeachingApi/GetIntoTeachingApi.csproj
+++ b/GetIntoTeachingApi/GetIntoTeachingApi.csproj
@@ -48,6 +48,7 @@
 		  <PrivateAssets>all</PrivateAssets>
 		</PackageReference>
 		<PackageReference Include="PropertyChanged.Fody" Version="3.3.2" PrivateAssets="All" />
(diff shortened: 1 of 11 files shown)


peteryates (approved):
Nice. Super-readable tests.

PR #544, opened 2021-03-12
Serialize attribute names to camelCase instead of PascalCase

ethax-ross:
By default `System.Text.Json` serializes attribute names to `camelCase` and `Newtonsoft.Json` will match the casing of the attribute being serialized (`PascalCase` for C# attributes).

Making these consistent will mean that we only have to check for property names in a single case when redacting sensitive data (otherwise we would have to check `firstName` and `FirstName`, for example).

`Newtonsoft.Json` will deserialize both camel and pascal case by default, so the existing jobs in the queue will still deserialize and process fine.

Fixes a typo in `Se/DeserializeChangeTracked` extension methods.

Extracts Json serialization tests out of `BaseModel` and into `JsonExtensionsTests`, which makes more sense.

### DIFF
--- a/GetIntoTeachingApi/Controllers/MailingListController.cs
+++ b/GetIntoTeachingApi/Controllers/MailingListController.cs
@@ -59,7 +59,7 @@ namespace GetIntoTeachingApi.Controllers
                 return BadRequest(this.ModelState);
             }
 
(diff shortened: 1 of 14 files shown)


peteryates (approved):
Makes sense

PR #545, opened 2021-03-12
Attach correlation IDs to logs

ethax-ross:
We want to be able to track a request from a client as it moves through the API and various log messages are created.

Add `CorrelationId` library that will automatically consume an incoming request ID (from the `X-Request-ID` header, which is the Rails default) and attach it to the scope of any log messages that the API writes.

If an `X-Request-Id` is not present then the library will generate a new one in the form of a GUID.


peteryates (approved):
Seems sensible, CorrelationId looks like an interesting approach

PR #550, opened 2021-03-17
Add mechanism to pause/resume the CRM integration

ethax-ross:
[Trello-1426](https://trello.com/c/ByTdLhsq/1426-add-endpoint-for-crm-team-to-call-when-they-are-doing-an-update-with-downtime)

The CRM is often taken offline for a short time when the CRM team perform a production update. This results in a flurry of errors from the API that - whilst not effecting end users - is creating noise in our error monitoring for something we can do nothing about.

To avoid this we want to add a mechanism for the CRM team to inform the API that they are taking the CRM offline. The API can then avoid calling the CRM whilst its unavailable and the errors will stop.

- Add AppSettings model

We want to be able to flag the CRM as unavailable when it has been intentionally taken offline by the CRM team (for an update, for example).

Storing this in Postgres felt like overkill, so there is now a simple `AppSettings` model that is backed by Redis. It has a single writable attribute to set how long we want to pause the CRM integration for (this is a `DateTime` so that we can auto-resume the integration and avoid potential user-error).

The idea is that the CRM will call an endpoint in the API and flag itself as being unavailable before an update is carried out. We can then prevent any calls to the CRM, which would otherwise result in errors.

- Add CRM client

Add a client for the CRM; this will enable the CRM team to call specific operations endpoints that pause/resume the API's integration with the CRM.

- Allow CRM integration to be paused/resumed

Adds two new operations endpoints for pausing/resuming the integration with the CRM.

This could have been more rest-ful by accepting a payload with the integration date, but I imagine if we get more `AppSettings` that can be modified externally we will want tight access control, which is currently at the routing
layer so I think having individual endpoints is the way to go for now.

Updates the health check operation to return `integration-paused` as the CRM status if applicable (the overall state will also be moved to `degraded`).

- Don't call CRM if the integration is paused

If we know the CRM is unavailable we should not call out to it (as it results in an error we can do nothing about).

Updates to prevent access tokens being generated when the CRM is offline (instead we return `404` not found which results in the user being treated as a new candidate/mirroring the current behavior sans error).

It is still possible for the API to call out to the CRM as part of the token verification process if:

1. The user generates a token
2. The CRM goes offline
3. The user attempts to verify the token

This should be unlikely to happen often and, whilst we could potentially handle it gracefully, its going to be a fair amount of work. I think we should see if/how often it occurs before committing to handling it explicitly.

- Skip sync/magic link generation jobs when CRM integration paused

When the CRM is unavailable we are unable to perform a sync/generate magic link tokens.

Ideally we would prevent the jobs from being picked up at all when the API is in this state, but Hangfire doesn't support this out of the box and the workarounds aren't ideal. It also means we don't have to monitor the `AppSettings` state to know when to resume processing jobs.

If the CRM is offline these jobs succeed immediately and log out that they have been effectively 'skipped'.

- Abort UpsertCandidateJob when CRM integration paused

Ideally we would fail the job silently or prevent it from being picked up by pausing the queue, but Hangfire does not support either of these.

Instead, we raise an error with a clear explanation so when we are alerted we don't have to investigate the source of the error.

### DIFF
--- a/GetIntoTeachingApi/Controllers/OperationsController.cs
+++ b/GetIntoTeachingApi/Controllers/OperationsController.cs
@@ -1,9 +1,11 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using GetIntoTeachingApi.Models;
 using GetIntoTeachingApi.Services;
 using GetIntoTeachingApi.Utils;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Swashbuckle.AspNetCore.Annotations;
@@ -14,12 +16,16 @@ namespace GetIntoTeachingApi.Controllers
     [ApiController]
     public class OperationsController : ControllerBase
     {
+        // Must be a substantial amount longer than the max expected offline duration
+        // of the CRM and less than 24 hours (the point at which we auto-purge any held PII).
+        private static readonly TimeSpan CrmIntegrationAutoResumeInterval = TimeSpan.FromHours(6);
         private readonly IStore _store;
         private readonly ICrmService _crm;
         private readonly INotifyService _notifyService;
         private readonly IHangfireService _hangfire;
         private readonly IRedisService _redis;
         private readonly IEnv _env;
+        private readonly IAppSettings _appSettings;
 
         public OperationsController(
             ICrmService crm,
@@ -27,7 +33,8 @@ namespace GetIntoTeachingApi.Controllers
             INotifyService notifyService,
             IHangfireService hangfire,
             IRedisService redis,
-            IEnv env)
+            IEnv env,
+            IAppSettings appSettings)
         {
             _store = store;
             _crm = crm;
@@ -35,6 +42,7 @@ namespace GetIntoTeachingApi.Controllers
             _hangfire = hangfire;
             _redis = redis;
             _env = env;
+            _appSettings = appSettings;
         }
 
         [HttpGet]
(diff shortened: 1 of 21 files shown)


GetIntoTeachingApi/Controllers/OperationsController.cs
```diff
@@ -77,5 +82,38 @@ public async Task<IActionResult> HealthCheck()
...
+            Summary = "Temporarily pauses the integration with the CRM.",
+            Description = "The CRM is taken offline for updates occasionally; this can result " +
+            "in errors when the API attempts to call out to the CRM. The CRM can call this endpoint " +
+            "to pause the API -> CRM integration (if not manually resumed it will auto-resume in 6 hours).",
+            OperationId = "PauseCrmIntegration",
+            Tags = new[] { "Operations" })]
+        [ProducesResponseType(typeof(HealthCheckResponse), StatusCodes.Status204NoContent)]
+        public IActionResult PauseCrmIntegration()
+        {
+            _appSettings.CrmIntegrationPausedUntil = DateTime.UtcNow.AddHours(6);
```
misaka, line 99: Would it be worth using an env var for `6` so that it can be adjusted as necessary?

ethax-ross: I think if it was likely to be changed then that would make sense, but I see it as something that should be static; essentially a value that is a fair bit greater than the maximum expected offline time for the CRM during an update and less than 24 hours (the maximum time we can hold on to PII for before we auto-purge it) -- I'll add a comment to make that clearer. It's also a safety net that hopefully won't come into effect if the CRM team follow protocol and call `/operations/resume_crm_integration` when the update is finished.

misaka: Sure. I was also thinking about testing environments where it could be useful to lower it, but if you reckon there's no need that's fine. 👍

jkempster34 (approved):

jkempster34 (approved):

PR #551, opened 2021-03-22
Update Gov UK Notify library

ethax-ross:
The library was previously hosted on bintray.com, but that will be deprecated on May 1st and the team have moved the latest version to nuget.org.

Update to use latest version on nuget.org.

I got a build warning due to GovUKNotify depending on JWT >= 7.0.0 and < 8.0.0, but v7.0.0 was not found (as there is no v7.0.0) - installing an applicable version manually resolves the warning.


peteryates (approved):
Looks good

PR #555, opened 2021-03-23
Update teaching event

jkempster34:
A new form will be added to the app to add teaching events. As part of this, teaching events should be able to be updated, so that their `status` can be changed.

This PR:
- Adds a new status to teaching event
- Updates the name and description of the teaching event add/update endpoint
- Update teaching event add/update endpoint to immediately cache building so that it can be retrieved straight away before sync
- Saves building in CRM separately so that it returns an ID.
- Caches event and building, so that they are immediately available

### DIFF
--- a/GetIntoTeachingApi/Controllers/TeachingEventsController.cs
+++ b/GetIntoTeachingApi/Controllers/TeachingEventsController.cs
@@ -182,8 +182,7 @@ namespace GetIntoTeachingApi.Controllers
 
             return CreatedAtAction(
                 actionName: nameof(Get),
-                controllerName: "TeachingEvents",
-                routeValues: new { id = teachingEvent.ReadableId },
+                routeValues: new { readableId = teachingEvent.ReadableId },
                 value: teachingEvent);
         }
 
--- a/GetIntoTeachingApi/Controllers/TeachingEventsController.cs
+++ b/GetIntoTeachingApi/Controllers/TeachingEventsController.cs
@@ -169,11 +169,11 @@ namespace GetIntoTeachingApi.Controllers
             Summary = "Adds or updates a teaching event.",
             Description = "If the `id` is specified then the existing teaching event will be " +
                           "updated, otherwise a new teaching event will be created.",
-            OperationId = "AddOrUpdateTeachingEvent",
+            OperationId = "UpsertTeachingEvent",
             Tags = new[] { "Teaching Events" })]
         [ProducesResponseType(typeof(TeachingEvent), StatusCodes.Status201Created)]
         [ProducesResponseType(typeof(IDictionary<string, string>), StatusCodes.Status400BadRequest)]
-        public async Task<IActionResult> AddOrUpdateTeachingEventAsync([FromBody] TeachingEvent teachingEvent)
+        public async Task<IActionResult> Upsert([FromBody] TeachingEvent teachingEvent)
         {
             if (!ModelState.IsValid)
             {
@@ -183,10 +183,12 @@ namespace GetIntoTeachingApi.Controllers
             if (teachingEvent.Building != null)
             {
                 _crm.Save(teachingEvent.Building);
-                await _store.SaveAsync(teachingEvent.Building);
             }
 
             _crm.Save(teachingEvent);
+
+            // Make the teaching event/building immediately available in the cache
+            await _store.SaveAsync(teachingEvent);
 
             return CreatedAtAction(
                 actionName: nameof(Get),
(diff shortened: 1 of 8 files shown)


GetIntoTeachingApi/Controllers/TeachingEventsController.cs
```diff
@@ -163,6 +163,37 @@ public async Task<IActionResult> Get([FromRoute, SwaggerParameter("The `readable
...
+        [Route("")]
+        [SwaggerOperation(
+            Summary = "Adds or updates a teaching event.",
+            Description = "If the `id` is specified then the existing teaching event will be " +
+                          "updated, otherwise a new teaching event will be created.",
+            OperationId = "AddOrUpdateTeachingEvent",
+            Tags = new[] { "Teaching Events" })]
+        [ProducesResponseType(typeof(TeachingEvent), StatusCodes.Status201Created)]
+        [ProducesResponseType(typeof(IDictionary<string, string>), StatusCodes.Status400BadRequest)]
+        public async Task<IActionResult> AddOrUpdateTeachingEventAsync([FromBody] TeachingEvent teachingEvent)
```
ethax-ross, line 176: Can we call thIs method `Upsert()`? I use that terminology in the candidate job and should probably start changing the others to match, so `UpsertMember` and `UpsertAttendee` etc. Then the operation id can be `UpsertTeachingEvent`.

GetIntoTeachingApi/Controllers/TeachingEventsController.cs
```diff
@@ -163,6 +163,37 @@ public async Task<IActionResult> Get([FromRoute, SwaggerParameter("The `readable
...
+                return BadRequest(ModelState);
+            }
+
+            if (teachingEvent.Building != null)
+            {
+                _crm.Save(teachingEvent.Building);
+                await _store.SaveAsync(teachingEvent.Building);
+            }
+
+            _crm.Save(teachingEvent);
```
ethax-ross, line 189: The teaching event isn't persisted to the store here, just the building? I think the building will be persisted as part of the event if you do something like (also adds a comment as to why we persist to the store because I don't think its necessarily obvious):

```
if (teachingEvent.Building != null)
{
  _crm.Save(teachingEvent.Building);
}

_crm.Save(teachingEvent);
// Make the event/building immediately available in the cache.
_store.SaveAsync(teachingEvent);
```

GetIntoTeachingApiTests/Controllers/TeachingEventsControllerTests.cs
```diff
@@ -191,6 +192,51 @@ public void ExchangeAccessTokenForAttendee_MissingCandidate_RespondsWithUnauthor
...
+
+            var response = await _controller.AddOrUpdateTeachingEventAsync(request);
+
+            var badRequest = response.Should().BeOfType<BadRequestObjectResult>().Subject;
+            var errors = badRequest.Value.Should().BeOfType<SerializableError>().Subject;
+            errors.Should().ContainKey(expectedErrorKey).WhichValue.Should().BeOfType<string[]>().Which.Should().Contain(expectedErrorMessage);
+        }
+
+        [Fact]
+        public async Task AddTeachingEvent_WhenRequestIsValid_SavesInCrmAndReturnsCreatedEvent()
```
ethax-ross, line 211: I would probably just have two 'valid' specs here, one to cover saving with a building and one saving without as that's the branch condition in your code (both can check the returned object is correct etc):

```
AddTeachingEvent_ValidRequestWithBuilding_SavesInCrmAndCaches()
AddTeachingEvent_ValidRequestWithoutBuilding_SavesInCrmAndCaches()
```

GetIntoTeachingApiTests/Services/StoreTests.cs
```diff
@@ -532,6 +532,26 @@ public async Task GetTeachingEventBuildings_ReturnsAll()
             var result = _store.GetTeachingEventBuildings().ToList();
 
             result.Should().HaveCount(5);
+        }
+
+        [Fact]
+        public async Task SaveAsync_WillAddAndSaveGivenEntity()
```
ethax-ross, line 538: I would avoid the term `entity` here as that's used to reference the Dynamics entities. If you go with the suggestion above of saving the teaching event (and implicitly the building) I would maybe change this to be something like:

```
public async Task SaveAsync_WithModelAndRelatedModel_PersistsBoth()
```

ethax-ross (commented):
Couple very minor suggestions, looking good overall

ethax-ross (approved):
LGTM

PR #558, opened 2021-03-25
Handle non-unique partial URL

jkempster34:
Add model error to the API response when the CRM has a validation error for non-unique partial URL

### DIFF
--- a/GetIntoTeachingApi/Models/Validators/TeachingEventValidator.cs
+++ b/GetIntoTeachingApi/Models/Validators/TeachingEventValidator.cs
@@ -1,18 +1,32 @@
 ﻿using System;
 using FluentValidation;
 using FluentValidation.Validators;
+using GetIntoTeachingApi.Services;
 
 namespace GetIntoTeachingApi.Models.Validators
 {
     public class TeachingEventValidator : AbstractValidator<TeachingEvent>
     {
-        public TeachingEventValidator()
+        private readonly ICrmService _crm;
+
+        public TeachingEventValidator(ICrmService crm)
         {
-            RuleFor(teachingEvent => teachingEvent.ReadableId).NotEmpty();
+            _crm = crm;
+
+            RuleFor(teachingEvent => teachingEvent.ReadableId)
+                .NotEmpty()
+                .Must(BeUnique)
+                .When(teachingEvent => teachingEvent.Id == null)
+                .WithMessage("Must be unique");
             RuleFor(teachingEvent => teachingEvent.Name).NotEmpty();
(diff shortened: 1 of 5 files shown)


GetIntoTeachingApi/Controllers/TeachingEventsController.cs
```diff
@@ -185,7 +186,22 @@ public async Task<IActionResult> Upsert([FromBody] TeachingEvent teachingEvent)
                 _crm.Save(teachingEvent.Building);
             }
 
-            _crm.Save(teachingEvent);
+            try
```
ethax-ross, line 189: Using exceptions to control flow like this is a smell... I would prefer the logic to live inside the validator, you're shoe-horning the `ModelState` manually here which is a good indicator. I would suggest adding a rule against the `ReadableId` that goes off to the CRM with a method like:

```
public TeachingEvent GetTeachingEvent(string readableId) {}
```

If the CRM returns a matching event then fail the validation.

GetIntoTeachingApi/GetIntoTeachingApi.csproj
```diff
@@ -54,7 +54,6 @@
 
 	<ItemGroup>
 	  <Folder Include="Redis\" />
-	  <Folder Include="Validators\" />
```
ethax-ross, line 57: Not sure this should have changed?

GetIntoTeachingApi/Models/Validators/TeachingEventValidator.cs
```diff
@@ -1,18 +1,32 @@
...
+        private readonly ICrmService _crm;
+
+        public TeachingEventValidator(ICrmService crm)
         {
-            RuleFor(teachingEvent => teachingEvent.ReadableId).NotEmpty();
+            _crm = crm;
+
+            RuleFor(teachingEvent => teachingEvent.ReadableId)
+                .NotEmpty()
+                .Must(IsPartialUrlUnique)
```
ethax-ross, line 18: I think the convention here is to use `Be...`, so `BeUnique`

GetIntoTeachingApi/Models/Validators/TeachingEventValidator.cs
```diff
@@ -1,18 +1,32 @@
...
+                .When(teachingEvent => teachingEvent.Id == null)
+                .WithMessage("External event title must be unique");
```
ethax-ross, line 20: The field name will be part of the message, so I think this would read better just as `Must be unique`

ethax-ross (approved):
LGTM

PR #559, opened 2021-03-26
Upgrade Terraform

sfawcett123:
### Task

The version of terraform is now version 14.9 it is time to switch to this version, as we also want to make changes in underlying modules

### Impact
No impact on the application

### DIFF
--- a/terraform/paas/provider.tf
+++ b/terraform/paas/provider.tf
@@ -14,7 +14,6 @@ locals {
 }
 
 provider "azurerm" {
-  version                    = ">= 2.0"
   skip_provider_registration = true
   features {}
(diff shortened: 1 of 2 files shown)


terraform/paas/provider.tf
```diff
@@ -14,7 +14,6 @@ locals {
 }
 
 provider "azurerm" {
-  version                    = ">= 2.0"
```
saliceti, line 17: Why remove this line? Seems like a good practice to pin the provider version to a minimum version

sfawcett123: Deprecated - showing warnings

PR #560, opened 2021-03-29
Add initial support for contract testing

ethax-ross:
### Trello card

### Context

We want to use a variation of 'contract testing' to achieve end-to-end integration tests across all services (Get an Adviser -> API -> CRM).

The API contract tests use the Get an Adviser payload snapshots as the test fixtures. These tests result in another set of request snapshots for the payloads sent to the CRM (which serve as the fixtures to the CRM contract tests).

A difficulty with these tests is ensuring that the services all have a consistent, global view of the service data state prior to executing the contract tests. Initially we will maintain the service data in `Contracts/Data` (to later be centralised in a GitHub repository, but duplicated in each service for now).

Eventually, the `Contracts/Output` will be 'published' to a separate GitHub repository, which will enable other services to pull in their test fixtures from the upstream service. This will enable us to develop features in each service independently and publish only when the change is ready to be reflected in another service.

- Add shared contract testing data

This data contains the application state for running the contract tests. It will be shared with the other applications (and eventually centralised so that they can easily stay in sync).

- Add input contracts from TTA

The `/Input` directory will contain the contracts output from the TTA service.

- Add contract test

Adds support for running contract tests output from the TTA service.

The tests run against a test instance of the application, which mocks out the CRM interactions and allows us to query objects that would have been sent to the CRM - they are serialized and saved as the output contracts of the API (for the CRM to ingest into their contract tests).

The application state is loaded prior to running tests, so that the data is consistent with what was available to the TTA service (and the current time is frozen consistently as well).

The contract output is committed and the snapshot is compared on subsequent runs. If the calls to the CRM change then the snapshot will differ and the test will fail (we would then commit the updated snapshot assuming the change is expected).

- Update README with contract test details

### Guidance to review

If the CRM team is happy with the output format then a centralised GitHub repository will be the next task so that we can 'publish' the contracts from each service and centralise the shared data state.


peteryates (approved):
This looks excellent (although I did just run through the tests and skipped the json).

Might be worth giving us a runthrough of the bigger picture at some point.

What do you see as being the process for when changes are made at the CRM side? Is updating the fixtures straightforward?

ethax-ross:
Cool, yeah happy to do a run through. If something changes its just a case of regenerating the contracts; which can be done in either the app if its a user-facing change (the Capybara specs will spit out new contracts in the form of requests to the API) and if its a change that just effects the API then we can just delete the old snapshots and re-run it, check the diff looks good and pass the new contracts onto the CRM team.

PR #561, opened 2021-03-29
Fix test environment for integration tests

ethax-ross:

To set up contract testing in the API we need to fix a couple of issues with the test environment when running integration tests:

- Don't run recurring jobs in test environment

We will be manually seeding the data in the test environment and not talking to the CRM, so halting the jobs makes sense.

- Point integration tests at per-test database

Previously the integration tests were pointing to `gis_test`, which is the template database we clone per-test.

Instead, we want the integration tests to point at the cloned database that is cleaned up once the test has finished.

- Fix flakey rate limit specs

Now the integration tests point at a per-test database it highlights a flakey spec. The rate limit tests were re-using the `WebApplicationFactory` across all tests by way of the `ClassFixture`. As our database is now setup/torn down per-test we need to recreate the factory as well, otherwise the factory is pointing to a destroyed database and we get an error (depending on if the tests run faster than the database is cleaned up or not, sometimes it would work other times it would fail).

### DIFF
--- a/GetIntoTeachingApi/Startup.cs
+++ b/GetIntoTeachingApi/Startup.cs
@@ -235,18 +235,22 @@ The GIT API aims to provide:
                 dbConfiguration.Migrate();
             }
 
-            // Configure recurring jobs.
-            const string everyFifthMinute = "*/5 * * * *";
-            const string everyMinute = "* * * * *";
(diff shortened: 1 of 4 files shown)


peteryates (approved):
Makes sense

PR #562, opened 2021-03-29
Make DateTime.UtcNow mockable

ethax-ross:

We reference `DateTime.UtcNow` in a number of places, primarily to default an attribute to the current UTC date/time. Up until now this has been fine and we can test the logic by matching these attributes with "a value that is close to DateTime.UtcNow".

Going forward we are snapshot matching in the contract tests, which requires the `DateTime` strings to be an exact match (or otherwise we would have to write custom code to process any `DateTime` attributes, which would end up being quite messy). The only way to reliably achieve this is to wrap `DateTime.UtcNow` in our own interface that can then be mocked out for the
tests.

This commit replaces (almost) all uses of `DateTime.UtcNow` with `IDateTimeProvider` - the couple of instances that haven't been updated won't effect the contract tests and are in difficult areas to change, due to the models being instantiated by the framework and not being able to provide an injected dependency easily.


peteryates (approved):
Nice.

PR #564, opened 2021-03-30
Add status filter to teaching event search

jkempster34:
Add status filter to teaching event search, default to event status open.

### DIFF
--- a/GetIntoTeachingApiTests/Services/StoreTests.cs
+++ b/GetIntoTeachingApiTests/Services/StoreTests.cs
@@ -71,7 +71,7 @@ namespace GetIntoTeachingApiTests.Services
 
             var ids = DbContext.TeachingEvents.Select(te => te.Id);
             ids.Should().BeEquivalentTo(mockTeachingEvents.Select(te => te.Id));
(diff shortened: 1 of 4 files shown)


GetIntoTeachingApiTests/Services/StoreTests.cs
```diff
@@ -504,6 +504,54 @@ public async void SearchTeachingEvents_FilteredByStartBefore_ReturnsMatching()
                 options => options.WithStrictOrdering());
         }
 
+        [Fact]
+        public async void SearchTeachingEvents_FilteredByDefaultStatusId_ReturnsOpenEvents()
+        {
+            await SeedMockTeachingEventsAndBuildingsAsync();
+            var request = new TeachingEventSearchRequest()
```
ethax-ross, line 511: This isn't really testing the default as you're explicitly setting the `StatusId` to `Open` - I think it should just be:

```
var request = new TeachingEventSearchRequest();
```

and then you could have a test in `TeachingEventSearchRequest` that checks the default value for the model

ethax-ross (approved):

PR #567, opened 2021-03-31
Include pending events in teaching event sync

jkempster34:
Pending teaching events must be included in the CRM sync else they will be removed as orphaned.


ethax-ross (approved):

GetIntoTeachingApiTests/Services/CrmServiceTests.cs
```diff
@@ -343,8 +347,8 @@ public void MatchCandidate_WithMasterChildRecords_ReturnsMasterRecord()
         {
```
jkempster34, line 347: Just noticed these whitespace changes sorry, I can remove them if you want. I bash Ctrl+k Ctrl+D (not sure what the mac equivalent is) to format a lot and it does this.

PR #574, opened 2021-04-06
Reduce job lock durations

ethax-ross:
We lock jobs that shouldn't run asynchronously; we've been using quite long lock timeouts and I believe this results in jobs (mainly the CrmSyncJob) taking forever to run locally after restarting your instance; the new job waits for the job that was killed to timeout. Reducing the timeout duration should speed things up locally and a 60s duration should be plenty for both jobs to complete (they both complete in under 10 seconds usually).


peteryates (approved):
Seems sensible.

PR #576, opened 2021-04-08
Remove event start at greater than now validation criteria

ethax-ross:
We want new events to be dated in the future, but existing/past events are still valid (as we show an archive of online events).

With this validation in place the event.StartAt was failing the validation criteria during mapping and being nullified, which results in the earliest `DateTime` value by default. We would be better checking that the type accepts a `null` value in the mapping operation to avoid this in the future.


peteryates (approved):
Makes sense!

PR #584, opened 2021-04-19
Only validate readableId of existing event when it has changed

jkempster34:
The `ReadableId` of a teaching event is validated to ensure that it is unique.

Currently, validation for the `ReadableId` of an existing event will be skipped because of the `When` check in the validator. This means that if the event is updated to have an existing `ReadableId` (in this situation the event will have a `Id` and non-unique `ReadableId`), then the event will try to save in the CRM and throw a `Microsoft.Xrm.Sdk.SaveChangesException`

This PR updates the validator to also validate events which do exist (have an `id`), but have a new `readableId`

### DIFF
--- a/GetIntoTeachingApiTests/Models/Validators/TeachingEventValidatorTests.cs
+++ b/GetIntoTeachingApiTests/Models/Validators/TeachingEventValidatorTests.cs
@@ -73,20 +73,6 @@ namespace GetIntoTeachingApiTests.Models.Validators
         }
 
         [Fact]
-        public void Validate_IdIsNotNullAndEventNotInStore_HasNoError()
-        {
-            var teachingEvent = new TeachingEvent() { Id = Guid.NewGuid(), ReadableId = "readable-id" };
-
-            _mockStore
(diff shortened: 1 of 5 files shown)


ethax-ross (approved):

GetIntoTeachingApiTests/Models/Validators/TeachingEventValidatorTests.cs
```diff
@@ -49,20 +51,53 @@ public void Validate_ReadableIdIsNullOrEmpty_HasError()
...
+                Id = Guid.NewGuid(),
+                ReadableId = "unique",
+            };
+
+            var existingTeachingEvent2 = new TeachingEvent
+            {
+                ReadableId = "not_unique",
+            };
+
+            _mockStore
```
ethax-ross, line 89: Shouldn't we be mocking the new store method here?

ethax-ross (commented):
Looks good just one comment where I think the wrong method is mocked

ethax-ross (approved):

PR #585, opened 2021-04-19
Add crm name teaching event

jkempster34:
Teaching Events have an `Event name` in the CRM (See left column of image below), and an `External event title` (`TeachingEvent.Name` in the API)

When adding a Teaching Event, the `Event name` should be set to equal the `External event title`. This will allow the CRM team to work with the events more easily (e.g. searching events will be easier)

image

### DIFF
--- a/GetIntoTeachingApi/Models/TeachingEvent.cs
+++ b/GetIntoTeachingApi/Models/TeachingEvent.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel.DataAnnotations.Schema;
 using System.Text.Json.Serialization;
 using FluentValidation;
 using GetIntoTeachingApi.Attributes;
@@ -11,6 +12,8 @@ namespace GetIntoTeachingApi.Models
     [Entity("msevtmgt_event")]
     public class TeachingEvent : BaseModel
     {
+        private string _name;
+
         public enum Status
         {
             Open = 222750000,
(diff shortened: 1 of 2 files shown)


GetIntoTeachingApi/Models/TeachingEvent.cs
```diff
@@ -41,6 +42,9 @@ public enum EventType
         public bool IsOnline { get; set; }
         [EntityField("dfe_externaleventtitle")]
         public string Name { get; set; }
+        [NotMapped]
```
ethax-ross, line 45: Can we also `JsonIgnore` this? no need for the clients to be aware of it

GetIntoTeachingApiTests/Models/TeachingEventTests.cs
```diff
@@ -82,5 +83,17 @@ public void IsInPerson_ReturnsCorrectly(bool isOnline, bool isVirtual, bool expe
...
+        [Fact]
+        public void CrmName_Get_ReturnsName()
+        {
+            const string name = "test";
+            var teachingEvent = new TeachingEvent()
+            {
+                Name = name,
+            };
+
+            teachingEvent.CrmName.Should().Be(name);
```
ethax-ross, line 96: Less cognitive load IMO:

```
var teachingEvent = new TeachingEvent() { Name = "name" }
teachingEvent.CrmName.Should().Be(teachingEvent.Name)
```

GetIntoTeachingApi/Models/TeachingEvent.cs
```diff
@@ -40,7 +43,20 @@ public enum EventType
...
+            {
+                _name = value;
+                CrmName = value;
+            }
+        }
+
+        [NotMapped]
+        [JsonIgnore]
+        [EntityField("msevtmgt_name")]
+        public string CrmName { get; set; }
```
ethax-ross, line 59: Can we rename this `InternalName`?

ethax-ross (approved):

PR #587, opened 2021-04-21
Add Teaching Event Building validator

jkempster34:
Teaching Event Buildings should always have a Venue name and a Postcode.


ethax-ross (approved):
LGTM - worth verifying no existing buildings fail validation during the CRM sync; if there are any we should ping the CRM team with details

ethax-ross (approved):

PR #593, opened 2021-04-22
Remove CrmETag

jkempster34:
A new ETag is generated for successful CRM syncs. This was implemented to prevent multiple requests for unchanged resources.

However, this was implemented at a time when the CRM sync ran every hour. Since then, it was requested [here](https://github.com/DFE-Digital/get-into-teaching-api/pull/268) that the CRM sync runs every 5 minutes instead.

This means that the ETag is no longer useful. Cache-Control is used as well (also with an expiration time of 5 minutes) that will ensure that requests are not made anyway.


ethax-ross (approved):
LGTM - when this goes out to prod we'll want to keep an eye on Grafana to make sure the API doesn't suddenly start receiving substantially more requests

PR #600, opened 2021-04-27
Make error response format consistent on manual validation

ethax-ross:
When we check for duplicate `ReadableId` values we end up returning a `BadRequest(ModelState)` if there is a duplicate, which results in a response in the format:

```
{
  "ReadableId": "'Readable Id' must be unique."
}
```

This is because the validation is performed manually/explicitly. Instead, we want to use the default error response format (which is used for all our other, automated validators by default) defined in an `InvalidModelStateResponseFactory` as part of the `ApiBehaviorOptions`:

```
{
  "type": "https://tools.ietf.org/html/rfc7231#section-6.5.1",
  "title": "One or more validation errors occurred.",
  "status": 400,
  "traceId": "|b0171620-4495860d4208836f.",
  "errors": {
    "ReadableId": [
      "'Readable Id' must be unique."
    ]
  }
}
```


peteryates (approved):
Makes sense

PR #605, opened 2021-04-29
Moving to GHCR from Docker hub

sfawcett123:
Dockerhub repository is now restrictive so moving to GHCR


ethax-ross (approved):
Looks fine, although I don't understand why there are changes to the `environment` keys?

PR #607, opened 2021-04-30
Allow overseas candidates to schedule a callback

ethax-ross:
We are re-enabling this feature in the TTA service.


sfawcett123 (approved):
LGTM

PR #608, opened 2021-05-04
Update read-only queries to non-tracking

jkempster34:
Using non-tracking queries for read-only scenarios can result in significant performance improvements.


ethax-ross (approved):
💯 💨

PR #611, opened 2021-05-06
Remove entity relationships

jkempster34:
Existing relationships do not get deleted when the related object is removed. 

Currently, related objects will be removed in the cache. However, because the relationship still exists, they will be added back when the CRM sync runs.

### DIFF
--- a/GetIntoTeachingApi/Models/BaseModel.cs
+++ b/GetIntoTeachingApi/Models/BaseModel.cs
@@ -262,28 +262,23 @@ namespace GetIntoTeachingApi.Models
 
                 if (attribute == null)
                 {
-
                     continue;
                 }
 
-                try
+                var relationship = new Relationship(attribute.Name);
+
+                if (value == null)
                 {
-                    if (value == null && source.RelatedEntities.Values.Count > 0)
+                    context.LoadProperty(source, relationship);
+
+                    if (source.RelatedEntities.ContainsKey(relationship))
                     {
-                        var relationship = new Relationship(attribute.Name);
-
                         Entity relatedEntity = source.RelatedEntities.Values.First().Entities.First();
-
                         crm.DeleteLink(source, relationship, target: relatedEntity, context);
-                        continue;
                     }
-                }
-                catch (Exception e)
-                {
-                    Console.WriteLine(e);
-                    throw;
-                }
 
+                    continue;
+                }
 
                 foreach (var relatedModel in EnumerableRelationshipModels(value))
                 {
@@ -294,7 +289,7 @@ namespace GetIntoTeachingApi.Models
                     }
 
                     var target = relatedModel.ToEntity(crm, context);
-                    crm.AddLink(source, new Relationship(attribute.Name), target, context);
+                    crm.AddLink(source, relationship, target, context);
                 }
             }
         }
--- a/GetIntoTeachingApi/Models/BaseModel.cs
+++ b/GetIntoTeachingApi/Models/BaseModel.cs
@@ -268,21 +268,13 @@ namespace GetIntoTeachingApi.Models
 
                 try
                 {
-                    if (value == null)
+                    if (value == null && source.RelatedEntities.Values.Count > 0)
                     {
-
                         var relationship = new Relationship(attribute.Name);
-                        context.LoadProperty(source, relationship);
 
+                        Entity relatedEntity = source.RelatedEntities.Values.First().Entities.First();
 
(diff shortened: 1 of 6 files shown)


GetIntoTeachingApi/Models/BaseModel.cs
```diff
@@ -260,8 +260,23 @@ private void MapRelationshipAttributesToEntity(Entity source, ICrmService crm, O
...
+                }
+
+                var relationship = new Relationship(attribute.Name);
+
+                if (value == null)
                 {
+                    context.LoadProperty(source, relationship);
+
+                    if (source.RelatedEntities.ContainsKey(relationship))
+                    {
```
jkempster34, line 275: This would only work when there can only be one related entity (event -> building)

GetIntoTeachingApi/Models/BaseModel.cs
```diff
@@ -260,8 +260,23 @@ private void MapRelationshipAttributesToEntity(Entity source, ICrmService crm, O
...
+                        Entity relatedEntity = source.RelatedEntities.First(entity => entity.Key == relationship).Value.Entities.First();
```
jkempster34, line 276: This would only work when there can only be one related entity (e.g. event -> building) 😑

jkempster34:
This is working with the below event. You can:
- Add event with building
- Remove the building (this will call `DeleteLink` so when the sync runs it won't get added again)
- Post the same event with no building again (will skip the call to `DeleteLink`)

```json
{
"typeId":222750009,
"statusId":222750003,
"readableId":"testing3",
"name":"TestingUpdated2",
"summary":"aa",
"description":"\u003cdiv\u003ea\u003cbr\u003e\u003ca href=\"https://www.google.com\"\u003elink to google\u003c/a\u003e\u003c/div\u003e",
"providerWebsiteUrl":"a",
"providerTargetAudience":"a",
"providerOrganiser":"a",
"providerContactEmail":"a@a.com",
"startAt":"2021-05-05T00:00:00+00:00",
"endAt":"2021-05-06T13:18:00+00:00",
"id":"fc528120-0c7d-4286-8c05-abccb582c7ad"
}
```


```json
{
"typeId":222750009,
"statusId":222750003,
"readableId":"testing3",
"name":"TestingUpdated2",
"summary":"aa",
"description":"\u003cdiv\u003ea\u003cbr\u003e\u003ca href=\"https://www.google.com\"\u003elink to google\u003c/a\u003e\u003c/div\u003e",
"providerWebsiteUrl":"a",
"providerTargetAudience":"a",
"providerOrganiser":"a",
"providerContactEmail":"a@a.com",
"startAt":"2021-05-05T00:00:00+00:00",
"endAt":"2021-05-06T13:18:00+00:00",
"id":"fc528120-0c7d-4286-8c05-abccb582c7ad",
"building":{
  "venue":"Joe venue Friday",
  "addressLine1":null,
  "addressLine2":null,
  "addressLine3":null,
  "addressCity":null,
  "addressPostcode":"M1 7AX",
  "imageUrl":null,
  "id":"018734aa-5cd6-4c10-8b0f-2b9059573c97"
  }
}
```

GetIntoTeachingApi/Models/BaseModel.cs
ethax-ross: So looking into this a bit the issue with doing it in the `BaseModel` is that we're not able to distinguish between all the possible scenarios. If you take, for example, a candidate record being saved with a single qualification that could either be interpreted as:

- Just add the single qualification
- Add the single qualification and remove all the other qualifications (as they're not present in our version of the relationship)

The only way to clear up the ambiguity would be to ensure that when you save a candidate the current state of the other qualifications is also present and up to date.

We would need to ensure this further up the chain, before the model mapping logic is called - probably as part of the `UpsertCandidateJob#SaveCandidate` - so load in the existing qualifications/past teaching positions here. We would then also need to query the current state of qualifications in the mapping logic so that we can do a diff and unlink any that are no longer present.

We would also need to check the change tracking as we do when we map an individual field attribute so that we can differentiate between `null` (all removed) and `not changed`.

I think this would be overcomplicated and difficult to maintain. A simpler approach may be to hook into the mapping process from the `TeachingEvent` class and perform a check/unlink the relationship there when its being mapped:

```
public bool ShouldMap(ICrmService crm, Context context) {
  if (buildingRemoved)
  {
    crm.RemoveLink(..., context)
  }

  return true;
}
```

PR #622, opened 2021-05-12
Use lookup() to extract secrets

sfawcett123:
If the secret is not in the map, then  secret["KEY"] will fail this can be resolved by using lookup( secret , "KEY" , default )


ethax-ross (approved):

peteryates (approved):
Makes sense

PR #625, opened 2021-05-13
Remove the `Telephone` field from request models

ethax-ross:
We have updated the clients to reference the new `AddressTelephone` field, so the original `Telephone` field can now be removed.


leoapost (approved):
LGTM

PR #627, opened 2021-05-14
Add initial support for Schools Experience service

ethax-ross:
[Trello-66](https://trello.com/c/ETWHyQJu/66-implement-api-endpoints-for-se)

This is the first pass at adding support for the Schools Experience service to the API; it will likely change as we start to integrate the API client into the service, but it should cover the main scenarios and data set used by SE. Best reviewed by-commit.

- Add Schools Experience fields to Candidate model

For the most part the Schools Experience application uses a subset of the fields already used by the Get into Teaching apps.

There are, however, a small number of new fields that needed adding:

```
- AddressLine3
- AddressStateOrProvince
- SecondaryEmail
- Telephone
- SecondaryTelephone
- MobileTelephone
- HasDbsCertificate
- DbsCertificateIssuedAt
- SecondaryPreferredTeachingSubjectId
```

- Support matchback Schools Experience requests

Add a new `SchoolExperienceSignUp` and associated validator to model the main transaction payload for signing up to the Schools Experience service.

Update rate limiting to ensure the sign up endpoint can only receive at most 250 requests/min.

- Add modelling for classroom experience

A `Candidate` is able to have associated classroom experience notes with their `contact` record. This is stored as a single field in the CRM on the `contact` entity, however we abstract this from the clients with the `ClassroomExperienceNote` model. It accepts the required parameters and has a `ToString()` method to format it appropriately for the underlying field.

Add string extension to nullify an empty string; we make use of this in `Candidate.FullName` to avoid it being included in the change tracking when not set (previously it was an empty string which would count as a change on the computed property).

- Add endpoint to associate classroom experience notes with candidates

Add an endpoint that will allow POSTing of new candidate experience notes against a given candidate.

- Remove default values from Candidate model

These were specific to the GiT website/TTA services and may not apply to Schools Experience, so they have been deferred to the request models.

| Endpoints      | Data Structures |
| ----------- | ----------- |
| Screenshot 2021-05-17 at 08 09 00 | Screenshot 2021-05-17 at 08 09 49 |

### DIFF
--- a/GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
+++ b/GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
@@ -1,0 +1,93 @@
+﻿using System;
+using System.Collections.Generic;
+using GetIntoTeachingApi.Jobs;
+using GetIntoTeachingApi.Models;
+using GetIntoTeachingApi.Services;
+using GetIntoTeachingApi.Utils;
+using Hangfire;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Swashbuckle.AspNetCore.Annotations;
+
+namespace GetIntoTeachingApi.Controllers.SchoolsExperience
+{
+    [Route("api/schools_experience/candidates")]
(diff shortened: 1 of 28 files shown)


peteryates (approved):
All looks sensible to me

PR #628, opened 2021-05-18
Make LINQ statements more efficient

jkempster34:
Visual Studio is complaining that these LINQ queries could be more efficient. 

If querying the context `FirstOrDefault` will always hit the database, whereas `Find` will first check whether the required entity is already loaded.

If querying a List or Array, `FirstOrDefault` uses a `for loop` vs `Find` which uses `foreach`, which is slower


jkempster34:
Closed because `Array.Find` is more difficult to read than what is already there

PR #630, opened 2021-05-19
Minor tweaks for Schools Experience

ethax-ross:
- Default country to UK for SE sign up

We only appear to accept sign ups to the Schools Experience service from UK candidates. Currently the country is set as an environment variable in the Rails app, but its easier to default this in the API.

- Default secondary email/telephone on SE sign up

The Rails app has logic in it to default the secondary telephone/email when pre-filling for form; this replicates the logic in the API so we can remove it from the app.

- Remove validation on DBS certificate date

In the SE app the DBS certificate issued at date is cleared when the `HasDbsCertificate` field is changed; this means it could be `true` and not have a date of issue, so we should remove this validation check.


peteryates (approved):
Nice. The API definitely feels like the better place to sort out the CRM 'oddities' like the email thing

PR #632, opened 2021-05-19
Remove teaching event relationships

jkempster34:
Existing relationships between `TeachingEvent` and `TeachingEventBuilding` do not get removed when the related Building is deleted. Currently, the Event is persisted in the cache with a null Building, which will then be re-added when the CRM sync runs.

Call `DeleteLink` in this scenario to remove the relationship in the CRM as well.

### DIFF
--- a/GetIntoTeachingApi/Models/TeachingEvent.cs
+++ b/GetIntoTeachingApi/Models/TeachingEvent.cs
@@ -1,10 +1,13 @@
 ﻿using System;
 using System.ComponentModel.DataAnnotations.Schema;
+using System.Linq;
+using System.Reflection;
 using System.Text.Json.Serialization;
 using FluentValidation;
 using GetIntoTeachingApi.Attributes;
 using GetIntoTeachingApi.Services;
 using Microsoft.Xrm.Sdk;
+using Microsoft.Xrm.Sdk.Client;
 using Swashbuckle.AspNetCore.Annotations;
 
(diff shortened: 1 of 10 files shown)


GetIntoTeachingApi/Models/TeachingEvent.cs
```diff
@@ -110,5 +113,40 @@ public TeachingEvent(Entity entity, ICrmService crm, IValidatorFactory validator
             : base(entity, crm, validatorFactory)
         {
         }
+
+        protected override bool ShouldMap(ICrmService crm, OrganizationServiceContext context)
+        {
+            if (Building == null)
```
ethax-ross, line 119: This feels fairly complex - is there any way we can simplify it/centralise some of this logic so it could be reused? I would expect the operation to be along the lines of:

```
var existingEvent = crm.GetTeachingEvent(Id);
var removeBuilding = Building == null && existingEvent.Building != null;

if (removeBuilding) {
  // one-liner here to unlink the building, something like:
  Unlink("Building", existingEvent.Building)
}
```

The `BaseModel` could then do the heavy-lifting by inspecting the annotations on the attribute passed to infer the entity logic names etc and ultimately call `crm.DeleteLink`. I don't think you'd need the call to `context.SaveChanges()` either, as we're using the `context` passed in the call to save can be deferred (it'll be called later on once the mapping is complete anyway).

PR #637, opened 2021-05-20
Remove teaching event relationships

jkempster34:
Existing relationships between TeachingEvent and TeachingEventBuilding do not get removed when the related Building is deleted. Currently, the Event is persisted in the cache with a null Building, which will then be re-added when the CRM sync runs.

Call DeleteLink in this scenario to remove the relationship in the CRM as well.

### DIFF
--- a/GetIntoTeachingApi/Models/BaseModel.cs
+++ b/GetIntoTeachingApi/Models/BaseModel.cs
@@ -83,6 +83,7 @@ namespace GetIntoTeachingApi.Models
             var entity = crm.MappableEntity(LogicalName(GetType()), Id, context);
             MapFieldAttributesToEntity(entity);
             MapRelationshipAttributesToEntity(entity, crm, context);
+            FinaliseEntity(entity, crm, context);
             return entity;
         }
 
@@ -108,6 +109,11 @@ namespace GetIntoTeachingApi.Models
             return true;
         }
 
+        protected virtual void FinaliseEntity(Entity source, ICrmService crm, OrganizationServiceContext context)
+        {
+            // Hook.
+        }
+
         protected void NotifyPropertyChanged([CallerMemberName] string propertyName = "")
         {
             var excluded = _propertyNamesExcludedFromChangeTracking.Any(p => p == propertyName);
@@ -116,6 +122,14 @@ namespace GetIntoTeachingApi.Models
             {
                 ChangedPropertyNames.Add(propertyName);
             }
+        }
+
+        protected void DeleteLink(Entity source, ICrmService crm, OrganizationServiceContext context, BaseModel modelToRemove, string modelToRemovePropertyName)
+        {
+            PropertyInfo property = GetType().GetProperty(modelToRemovePropertyName);
+            var attribute = EntityRelationshipAttribute(property);
+            var target = modelToRemove.ToEntity(crm, context);
+            crm.DeleteLink(source, new Relationship(attribute.Name), target, context);
         }
 
         private static IList NewListOfType(Type type)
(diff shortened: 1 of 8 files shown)


GetIntoTeachingApi/Models/BaseModel.cs
```diff
@@ -118,6 +118,14 @@ protected void NotifyPropertyChanged([CallerMemberName] string propertyName = ""
             }
         }
 
+        protected void DeleteLink(ICrmService crm, OrganizationServiceContext context, BaseModel childToRemove, string childToRemovePropertyName)
```
ethax-ross, line 121: Can we rename to `modelToRemove` and `propertyNameContainingModelToRemove`? `child` implies a parent/child relationship which isn't always going to be the case

GetIntoTeachingApi/Models/BaseModel.cs
```diff
@@ -118,6 +118,14 @@ protected void NotifyPropertyChanged([CallerMemberName] string propertyName = ""
             }
         }
 
+        protected void DeleteLink(ICrmService crm, OrganizationServiceContext context, BaseModel childToRemove, string childToRemovePropertyName)
+        {
+            PropertyInfo property = GetType().GetProperty(childToRemovePropertyName);
+            var attribute = EntityRelationshipAttribute(property);
+            var target = childToRemove.ToEntity(crm, context);
+            crm.DeleteLink(MapToEntity(crm, context), new Relationship(attribute.Name), target, context);
```
ethax-ross, line 126: Re-mapping to an entity here feels off -- I wonder if instead of doing the logic in `ShouldMap` would we be better having another callback that fits in later, something like:

```
var entity = crm.MappableEntity(LogicalName(GetType()), Id, context);
MapFieldAttributesToEntity(entity);
MapRelationshipAttributesToEntity(entity, crm, context);
FinaliseEntity(crm, context);

...

public void FinaliseEntity(ICrmService crm, OrganizationServiceContext context)
{
...
DeleteLink(crm, context, existingEvent.Building, nameof(existingEvent.Building));
...
base.FinaliseEntity(crm, context);
}
```

Then you have access to the `source` already with `this`?

```
crm.DeleteLink(this, new Relationship(attribute.Name), context);
```

ethax-ross (commented):
Looks pretty close, just a couple of tweaks where I think we could simplify it a bit

GetIntoTeachingApiTests/Models/TeachingEventTests.cs
```diff
@@ -1,13 +1,27 @@
...
     public class TeachingEventTests
     {
+        private readonly OrganizationServiceContext _context;
+        private readonly Mock<TeachingEventBuilding> _building;
+        private readonly Mock<ICrmService> _mockCrm;
+
+        public TeachingEventTests()
+        {
+            _context = new OrganizationServiceContext(new Mock<IOrganizationService>().Object);
+            _building = new Mock<TeachingEventBuilding>();
```
ethax-ross, line 21: `_mockBuilding`

GetIntoTeachingApiTests/Services/CrmServiceTests.cs
```diff
@@ -517,6 +529,19 @@ public void GetTeachingEvent_WhenTeachingEventDoesNotExist_ReturnsNull()
...
+        [Fact]
+        public void GetTeachingEvent_CallLoadProperty()
+        {
+            _mockService.Setup(mock => mock.CreateQuery("msevtmgt_event", _context))
+                .Returns(MockTeachingEvents());
+
+            var result = _crm.GetTeachingEvent("event_one");
+
+            result.ReadableId.Should().Be("event_one");
+            _mockService.Verify(mock => mock.LoadProperty(It.IsAny<Entity>(),
```
ethax-ross, line 541: I'd be inclined to roll this into the existing `GetTeachingEvent_WhenTeachingEventExists_ReturnsTeachingEvent` test

ethax-ross (approved):
LGTM

PR #643, opened 2021-05-25
Add merged attribute to SchoolsExperienceSignUp

ethax-ross:
Indicates if the contact record has been merged and should have a `masterId`. I didn't think we'd need this as well as `masterId` but the client app references both, so putting it in to make the migration easier.


peteryates (approved):
Looks good

PR #649, opened 2021-06-01
Update CdsServiceClient

ethax-ross:
The CdsServiceClient has had a major version bump (we were a major version behind anyway) and the core classes have been renamed to fall inline with the Microsoft PowerPlatform "Dataverse" branding.

Functionality-wise the interfaces remain largely consistent, just a couple of minor updates to the configuration of the client and the method to retrieve the current user (performed in the health check).

I plan on pushing this to dev and doing some full end-to-end testing with each of the clients to ensure everything still works correctly.


jkempster34 (approved):
:muscle:

PR #651, opened 2021-06-02
Format international phone numbers

ethax-ross:
We are changing the front-end to include the country dial-in code in international phone numbers.

If a candidate enters an international number we want to:

- Remove any non-digit character
- Prefix the number with `00`
- If the number has the UK dial-in code, swap it for a `0`

Telephone numbers for UK-based candidates should remain unchanged.

Due to the format change we can no longer easily use the telephone number itself to determine the destination; instead, we will assume if the candidate chooses a non-UK country then the telephone number is international.

### DIFF
--- a/GetIntoTeachingApi/Models/TeacherTrainingAdviserSignUp.cs
+++ b/GetIntoTeachingApi/Models/TeacherTrainingAdviserSignUp.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Linq;
 using System.Text.Json.Serialization;
+using System.Text.RegularExpressions;
 using GetIntoTeachingApi.Services;
 using GetIntoTeachingApi.Utils;
 using Swashbuckle.AspNetCore.Annotations;
@@ -9,6 +10,8 @@ namespace GetIntoTeachingApi.Models
 {
     public class TeacherTrainingAdviserSignUp
     {
+        private string _addressTelephone;
+
         public Guid? CandidateId { get; set; }
         public Guid? QualificationId { get; set; }
         public Guid? SubjectTaughtId { get; set; }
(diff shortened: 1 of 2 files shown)


GetIntoTeachingApi/Models/TeacherTrainingAdviserSignUp.cs
```diff
@@ -164,6 +178,25 @@ private Candidate CreateCandidate()
...
+        {
+            var international = CountryId != LookupItem.UnitedKingdomCountryId;
+
+            if (international && telephone != null)
+            {
+                // Remove non-digit characters.
+                telephone = Regex.Replace(telephone, "[^0-9]", string.Empty);
+
+                // Prefix with 00.
+                telephone = $"00{telephone}";
```
peteryates, line 191: Is it safe to assume that the person who typed it didn't already prepend the exit code?

ethax-ross: Yeah we validate on the front-end to ensure the phone number starts with the country dial-in code

ethax-ross: I didn't know that was called an exit code though - I'll update the comment 😄

peteryates (commented):
Looks good, left one question

peteryates (approved):

GetIntoTeachingApi/Models/TeacherTrainingAdviserSignUp.cs
```diff
@@ -35,7 +38,18 @@ public class TeacherTrainingAdviserSignUp
         public DateTime? DateOfBirth { get; set; }
         public string TeacherId { get; set; }
         public string DegreeSubject { get; set; }
-        public string AddressTelephone { get; set; }
+        public string AddressTelephone
+        {
```
jkempster34, line 42: Is it worth using the arrow syntax to save some lines?

```c#
 get => return SanitizeTelephone(_addressTelephone);
 set => _addressTelephone = value;
```

ethax-ross: Yeah that's a good shout, I'll maybe see if I can get the linter setup to catch that

PR #660, opened 2021-06-08
Conditionally set teaching event sign up attributes

ethax-ross:
I've not got to the root cause of this bug, but if you sign up for a teaching event then perform a matchback and sign up for another teaching event without opting into the mailing list the API fails to send the sign up through to the CRM. The underlying exception is obscured by another exception raised by the ServiceClient library, so its not clear whats gone wrong.

The cause appears to be sending a `null` option set value for two of the optional properties; as these are optional we can instead not set them at all if they are passed to the API as null, which gets around the error for now. I'll look into raising an issue for the bug in the ServiceClient repo.


peteryates (approved):
Seems sensible

PR #663, opened 2021-06-08
Fix new event error

jkempster34:
We were seeing an error when posting a new event. Add some null checks and tests to fix.

### DIFF
--- a/GetIntoTeachingApi/Models/TeachingEvent.cs
+++ b/GetIntoTeachingApi/Models/TeachingEvent.cs
@@ -116,7 +116,7 @@ namespace GetIntoTeachingApi.Models
         {
             var existingEvent = crm.GetTeachingEvent(ReadableId);
 
-            bool removeBuilding = Building == null && existingEvent.Building != null;
+            bool removeBuilding = Building == null && existingEvent?.Building != null;
 
             if (removeBuilding)
             {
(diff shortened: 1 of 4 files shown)


GetIntoTeachingApi/Models/TeachingEvent.cs
```diff
@@ -116,11 +116,14 @@ protected override void FinaliseEntity(Entity source, ICrmService crm, Organizat
         {
             var existingEvent = crm.GetTeachingEvent(ReadableId);
 
-            bool removeBuilding = Building == null && existingEvent.Building != null;
-
-            if (removeBuilding)
+            if (existingEvent != null)
```
ethax-ross, line 119: Don't feel strongly but you could drop the nested `if`s by using a safe navigation here:

```
bool removeBuilding = Building == null && existingEvent?.Building != null;
```

GetIntoTeachingApiTests/Models/TeachingEventTests.cs
```diff
@@ -136,6 +136,25 @@ public void ToEntity_WhenBuildingIsRemoved_DeletesLink()
...
+
+            var newEvent = new TeachingEvent
+            {
+                ReadableId = "readableId",
+                Building = new TeachingEventBuilding()
+            };
+
+            var entity = newEvent.ToEntity(_mockCrm.Object, _context);
+
+            entity.Should().BeOfType<Entity>();
```
ethax-ross, line 155: I would maybe go with a not null check here -- the method is typed so it'll always return an `Entity`

ethax-ross (commented):
Looks good, couple minor comments

ethax-ross (approved):

PR #667, opened 2021-06-14
Split CandidateUpserter into multiple, atomic jobs

ethax-ross:
[Trello-1627](https://trello.com/c/3B2nts1i/1627-split-upsertcandidatejob)

- Add job to claim a callback booking quota slot

When someone books a callback we need to ensure the callback booking slot they select has its available slots decremented. Currently we do this in the `UpsertCandidateJob`, but this commit adds a new job to perform this function.

- Add generic UpsertModelJob

As we are splitting the `UpsertCandidateJob` into multiple independent jobs we need to be able to upsert a variety of records that have a `CandidateId` (are related to a candidate). This commit adds an `UpsertModelJob` that can take any `BaseModel` implementing `IHasCandidateId` and upsert it to the CRM.

Update applicable models to implement the new interface.

- Have CandidateUpserter queue jobs for sub-tasks

The main purpose of the `CandidateUpserter` is to upsert a `Candidate` record. As a by-product it also supports upserting all the objects related to a candidate, such as accepted privacy policy, qualifications, phone calls etc.

Currently this is all done in the `CandidateUpserter`, however as this is all wrapped in a single job if any part fails and it retries it could cause a duplicate record to be created.

To get around this, the `CandidateUpserter` is updated to queue more jobs for each upsert sub-task (once the candidate is successfully upserted). This way, if anything fails its retried independently and duplicates should be avoided.

Remove catch and re-raise; it doesn't seem to be necessary, Sentry is picking up the exceptions already.

### DIFF
--- a/GetIntoTeachingApi/Models/CandidatePastTeachingPosition.cs
+++ b/GetIntoTeachingApi/Models/CandidatePastTeachingPosition.cs
@@ -7,7 +7,7 @@ using Microsoft.Xrm.Sdk;
 namespace GetIntoTeachingApi.Models
 {
     [Entity("dfe_candidatepastteachingposition")]
-    public class CandidatePastTeachingPosition : BaseModel
+    public class CandidatePastTeachingPosition : BaseModel, IHasCandidateId
     {
         public enum EducationPhase
         {
--- a/GetIntoTeachingApiTests/Jobs/ClaimCallbackBookingSlotJobTests.cs
+++ b/GetIntoTeachingApiTests/Jobs/ClaimCallbackBookingSlotJobTests.cs
@@ -1,0 +1,103 @@
+﻿using System;
+using FluentAssertions;
+using GetIntoTeachingApi.Adapters;
+using GetIntoTeachingApi.Jobs;
+using GetIntoTeachingApi.Models;
+using GetIntoTeachingApi.Services;
+using GetIntoTeachingApi.Utils;
+using GetIntoTeachingApiTests.Helpers;
+using Microsoft.Extensions.Logging;
+using Moq;
+using Xunit;
+
+namespace GetIntoTeachingApiTests.Jobs
+{
+    public class ClaimCallbackBookingSlotJobTests
+    {
+        private readonly Mock<IPerformContextAdapter> _mockContext;
+        private readonly Mock<IAppSettings> _mockAppSettings;
+        private readonly Mock<ICrmService> _mockCrm;
+        private readonly DateTime _scheduledAt;
+        private readonly IMetricService _metrics;
+        private readonly ClaimCallbackBookingSlotJob _job;
+        private readonly Mock<ILogger<ClaimCallbackBookingSlotJob>> _mockLogger;
+
+        public ClaimCallbackBookingSlotJobTests()
+        {
+            _mockContext = new Mock<IPerformContextAdapter>();
+            _mockLogger = new Mock<ILogger<ClaimCallbackBookingSlotJob>>();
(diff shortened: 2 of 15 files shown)


GetIntoTeachingApi/Models/CandidatePastTeachingPosition.cs
```diff
@@ -7,7 +7,7 @@
 namespace GetIntoTeachingApi.Models
 {
     [Entity("dfe_candidatepastteachingposition")]
-    public class CandidatePastTeachingPosition : BaseModel
+    public class CandidatePastTeachingPosition : BaseModel, IHasCandidateId
```
jkempster34, line 10: Could the interface have a more descriptive name like `ICandidateModelUpsertable`? or would that not make sense?

GetIntoTeachingApiTests/Jobs/ClaimCallbackBookingSlotJobTests.cs
```diff
@@ -0,0 +1,103 @@
...
+            Action action = () => _job.Run(_scheduledAt, null);
+
+            action.Should().Throw<InvalidOperationException>()
+                .WithMessage("ClaimCallbackBookingSlotJob - Aborting (CRM integration paused).");
+        }
+
+        private bool IsMatch(object objectA, object objectB)
+        {
+            objectA.Should().BeEquivalentTo(objectB);
+            return true;
```
jkempster34, line 100: Can `IsMatch` return void? Or does that not work?

GetIntoTeachingApi/Models/CandidatePastTeachingPosition.cs
ethax-ross: Yeah I struggled with this interface name 😞  I settled on `IHasCandidateId` because it was the clearest/most accurate I could come up with; `ICandidateModelUpsertable` suggests if you implement the interface then your record will be upsertable, but that's not the case (it also needs to extend `BaseModel` and have appropriate data annotations)

GetIntoTeachingApiTests/Jobs/ClaimCallbackBookingSlotJobTests.cs
ethax-ross: It needs to evaluate to a boolean so the calling `It.Is` evaluates correctly; its wrapping a fluent assertion call as that was the easiest way to do an equivalency test that I could find (the objects need to be equal in attribute values but not the same instance of the object)

GetIntoTeachingApiTests/Services/CandidateUpserterTests.cs
```diff
@@ -100,30 +124,24 @@ public void Upsert_WithPrivacyPolicy_SavesPrivacyPolicy()
...
         }
 
-        private bool IsMatch(object objectA, object objectB)
+        private static bool IsMatch(object objectA, object objectB)
         {
             objectA.Should().BeEquivalentTo(objectB);
             return true;
         }
+
+        private static bool IsMatch<T>(T modelA, string modelBJson)
```
jkempster34, line 140: These helpers are static in this test class but not in the others. Not sure if it is a big deal

jkempster34 (approved):
LGTM 🥇

GetIntoTeachingApiTests/Services/CandidateUpserterTests.cs
ethax-ross: Yeah they could all be static, I'll change the other ones when I'm in there, or try and figure out a linter to sort this stuff 😅

PR #669, opened 2021-06-16
Optimize candidate match back queries

ethax-ross:
The candidate match back queries currently use LINQ, which is easier to write/understand/test than an QueryExpression, however the drawback is you can't easily scope the fields coming back from Dynamics to just those that you want (it is possible, but you need to pluck the fields you want in the `Select` clause, which doesn't fit well with how we instantiate our models from Dynamics entities).

Update match back queries to use `QueryExpression`, which should be faster albeit less readable/testable.


peteryates (approved):
Make sense. The Linq syntax is definitely nicer!

PR #677, opened 2021-06-24
Integrate with the Find & Apply candidates API

ethax-ross:
### Trello card

[Trello-1641](https://trello.com/c/LwotLDGF/1641-dev-find-apply-integration-work-with-the-api-crm)

### Context

The Find & Apply team are providing a candidates API that exposes candidate information updated since a given date/time. We want to query this API for the GiT API and sync up the `id` of their candidates with those matched in the CRM.

Initially we will only be matching back on email address and updating existing candidates (creating new candidates in the CRM that exist in Find & Apply but don't match back will come later).

We want to run the sync operation hourly and only in non-production environments. We will only be concerned with records updated _after_ the first sync run (past records will be bulk-imported by another process/job). 

- Store find/apply last sync date in Redis

We need to keep track of the date that the Find & Apply API was last synced successfully (as we pass this as a parameter in the subsequent API call).

Update `AppSettings` to enable storing the last sync date.

- Add find/apply API credentials

In order to access the Find & Apply API we will need the URL for the instance (sandbox or production) and a valid API key.

- Support match back on email address only

The normal match back process involves matching the email plus 2/3 of first name, last name and date of birth.

As the Find & Apply API is only going to be exposing the email address of the candidates initially we need to support matching back on just that field.

Replicates the current match back logic (taking into account status, duplicate score and date modified) and picks the first with the correct email.

- Add FindApplyId to Candidate model

When we sync with the Find & Apply API we will be storing their candidate id against the corresponding `contact` in the CRM; this adds the field mapping for the id.

- Model response from find/apply API

Add the models that map to the response from the Find & Apply candidates API.

- Add job to sync a find/apply candidate with the CRM

Add a job that takes a Find & Apply candidate model and matches it against a `contact` in the CRM. If there is a match the `FindApplyId` is set and the model is persisted back to the CRM.

- Add job to sync with find/apply API

The job requests Find & Apply candidates that have been updated since the last sync date. If no sync has been performed yet it will use the current date/time (we will pre-fill past records separately). A job is then queued for each candidate to sync it with the CRM.

Flurl is used for making the HTTP requests.

- Run find/apply sync job on a schedule

We want to sync with the Find & Apply API periodically; initially this will be hourly and only in our development and staging environments (we don't want this enabled in production yet).

### Guidance to review

Best reviewed by-commit.

### DIFF
--- a/GetIntoTeachingApi/GetIntoTeachingApi.csproj
+++ b/GetIntoTeachingApi/GetIntoTeachingApi.csproj
@@ -50,6 +50,7 @@
 		<PackageReference Include="GovukNotify" Version="4.0.0" />
 		<PackageReference Include="JWT" Version="7.3.1" />
 		<PackageReference Include="Microsoft.PowerPlatform.Dataverse.Client" Version="0.4.12" />
+		<PackageReference Include="Flurl.Http" Version="3.2.0" />
 	</ItemGroup>
 
 	<ItemGroup>
@@ -65,6 +66,8 @@
 	</ItemGroup>
 	<ItemGroup>
 	  <None Remove="Controllers\GetIntoTeaching\" />
+	  <None Remove="Flurl" />
+	  <None Remove="Flurl.Http" />
 	  <None Remove="Models\FindApply\" />
 	</ItemGroup>
 	<ItemGroup>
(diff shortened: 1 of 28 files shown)


jkempster34 (dismissed):
Had a thorough read of this. Only things I can think of, is that `Response` is usually outgoing and that I had some confusion between the two `Candidate` models. But Response and Candidate are both namespaced and make sense in their context so 🤷‍♂️ 

LGTM 💯 💯

ethax-ross:
Yeah I dithered a bit on the naming but I think ultimately its the core `Candidate` model that should be renamed to `Contact` to better reflect the CRM domain... I see what you mean about `Response`, I'll perhaps change that to `Result` or similar to avoid confusion, but I'll do that as part of the next PR.

jkempster34 (approved):

PR #678, opened 2021-06-25
Tidy upsert teaching event action

ethax-ross:
Minor refactoring to add an overload of the `SaveAsync` method to the store that accepts a single model and tidy up the logic in the `Upsert` action of the `TeachingEventsController`.


jkempster34 (approved):
Looks much neater 👍

PR #679, opened 2021-06-28
Configure DataProtection to persist in Redis

ethax-ross:
We don't directly use the `DataProtection` features of ASP.NET, but they are on by default and Hangfire appears to use them. We are seeing a number of anti-forgery check failures in the test environment, which appears to be due to running multiple instances and the `DataProtection` library persisting the tokens to disk by default. This means if a request goes to one  instance then the other it will fail the forgery check as the signatures wont match.

It doesn't appear to effect production as I believe its only the Hangfire dashboard that utilises the library (which is not enabled in prod).

Update to persist the keys in Redis, so that they are consistent across multiple instances.


jkempster34 (approved):
👍

PR #680, opened 2021-06-28
Only write FindApplyId attribute on sync job

ethax-ross:
Previously we were retrieving a matching `Candidate`, populating the `FindApplyId` and writing that back to the CRM. The issue with this is that the `Candidate` returned from the match query will be fully populated, so we end up writing back attributes needlessly.

Instead, we instantiate a new `Candidate` with the relevant `id` and populate the `FindApplyId`, ensuring that is the only attribute that will be written back to the CRM.


jkempster34 (approved):
💯

PR #685, opened 2021-06-29
Add basic auth to Hangfire dashboard

ethax-ross:
- Rename HangfireDashboardAuthorizationFilter

It wasn't clear from the name of the class that this is responsible for making the Hangfire dashboard only available in certain environments.

- Add basic auth to Hangfire dashboard

The Hangfire dashboard is only available in the dev/test environment, however there is no authentication in front of it at the moment. Whilst its not particularly exploitable, someone could re-queue a bunch of jobs and potentially clog up our workers.

Add basic authentication in front of the Hangfire dashboard.


leoapost (approved):
Looks good to me

PR #686, opened 2021-06-29
Clean up startup

jkempster34:


GetIntoTeachingApi/AppStart/DatabaseUtility.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            var dbContext = _serviceScope.ServiceProvider.GetService<GetIntoTeachingDbContext>();
+
+            // Initial CRM sync.
+            if (!dbContext.PickListItems.Any())
+            {
+                RecurringJob.Trigger(JobConfiguration.CrmSyncJobId);
+            }
+
+            // Initial locations sync.s
```
ethax-ross, line 39: Typo

GetIntoTeachingApi/AppStart/Prometheus/PrometheusMetricLabels.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using GetIntoTeachingApi.Utils;
+using Prometheus;
+
+namespace GetIntoTeachingApi.AppStart.Prometheus
+{
+    public static class PrometheusMetricLabels
```
ethax-ross, line 8: May be worth merging the Utils/ApplicationExtensions which does prometheus config into here as well

GetIntoTeachingApi/AppStart/StartupStaging.cs
```diff
@@ -0,0 +1,52 @@
+﻿using GetIntoTeachingApi.AppStart.Hangfire;
+using GetIntoTeachingApi.AppStart.Prometheus;
+using Hangfire;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace GetIntoTeachingApi.AppStart
+{
+    public class StartupStaging : Startup
```
ethax-ross, line 10: Is there much of a difference between staging and production? worth having staging inherit from production and make any further tweaks in the subclass if they are very similar?

GetIntoTeachingApi/Controllers/TeachingEventsController.cs
```diff
@@ -185,26 +184,41 @@ public async Task<IActionResult> Get([FromRoute, SwaggerParameter("The `readable
             [FromBody, SwaggerRequestBody("Teaching event to upsert.", Required = true)] TeachingEvent teachingEvent,
             [FromServices] IOptions<ApiBehaviorOptions> apiBehaviorOptions)
         {
-            ValidateForUpsertOperation(teachingEvent);
```
ethax-ross, line 188: Have these changes been reverted by a bad rebase?

ethax-ross (commented):
Definite improvement, we can maybe do a bit more to DRY up the staging startup I think though as its very similar to prod. Also I think a bunch of changes are in here that shouldn't be, not sure what's gone on there?

PR #687, opened 2021-06-29
Clean up startup

jkempster34:
The `Startup` class is getting a bit unwieldy. Update it to use extension/static classes to help keep things organised.

We have also investigated having one `Startup` per environment, however, that was quite hard to follow.


ethax-ross (commented):
I think I prefer this approach to the other one; it makes the single startup nice and simple/easier to read. I think its easier to get your head around than having the individual startups, even though that perhaps goes against convention.

GetIntoTeachingApi/AppStart/DatabaseUtility.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            var dbContext = scope.ServiceProvider.GetService<GetIntoTeachingDbContext>();
+
+            // Initial CRM sync.
+            if (!dbContext.PickListItems.Any())
+            {
+                RecurringJob.Trigger(JobConfiguration.CrmSyncJobId);
+            }
+
+            // Initial locations sync.s
```
ethax-ross, line 32: Typo

GetIntoTeachingApi/AppStart/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,168 @@
...
+                });
+
+                c.OperationFilter<AuthOperationFilter>();
+                c.EnableAnnotations();
+            });
+
+            services.AddFluentValidationRulesToSwagger();
+        }
+
+        public static void AddApiHangfire(this IServiceCollection services, IEnv env, bool useMemoryStorage)
```
ethax-ross, line 112: Just `AddHangfire`? not sure what the `Api` references?

GetIntoTeachingApi/AppStart/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,168 @@
...
+            services.AddSingleton<IRequestResponseLoggingConfiguration, RequestResponseLoggingConfiguration>();
+        }
+
+        public static void AddDatabase(this IServiceCollection services, IEnv env)
+        {
+            var connectionString = DbConfiguration.DatabaseConnectionString(env);
+            services.AddDbContext<GetIntoTeachingDbContext>(b => DbConfiguration.ConfigPostgres(connectionString, b));
+        }
+
+        public static void AddApiAuthentication(this IServiceCollection services)
```
ethax-ross, line 67: `AddApiClientAuthentication`?

GetIntoTeachingApi/AppStart/Startup.cs
```diff
@@ -0,0 +1,150 @@
...
+            }
+
+            app.UseAuthentication();
+
+            app.UseHttpsRedirection();
+
+            app.UseRequestResponseLogging();
+
+            app.ConfigureHangfire(
+                workerCount: !_env.IsDevelopment ? 20 : null,
```
ethax-ross, line 97: Reverse this to make it easier to read?

```
env.IsDevelopment ? null : 20;
```

ethax-ross (commented):
Couple minor comments but this is _way_ nicer 💯

jkempster34:
> Couple minor comments but this is _way_ nicer 💯

Thanks 🥇 All fixed.

ethax-ross (approved):
Looks good 👍

PR #690, opened 2021-06-30
Update to ASP.NET Core 5

ethax-ross:
- Upgrade to .NET Core 5/all dependencies

Upgrade the two project files to .NET Core 5 - whilst this version does not have LTS it will be supported until the end of next year, by which point we should have already upgraded to the next version with LTS; upgrading now should make that process easier (so we're not jumping from 3 all the way to 6).

Bring all dependencies up to the latest available versions.

- Update BaseModel validation

The latest version of FluentValidation dropped support for `Validate(model)` - instead, it expects a validation context and recommends using `new ValidationContext<object>(model)` to perform dynamic validation when the instance is not known ahead of time.

- Make property validators generic

The latest version of FluentValidation changed the property validator interface to accept generic types; this updates our property validators and all references.

- Change disposed context exception text

Later versions of EF Core throw a slightly different error message if you try and access a disposed DB context (which we do to test the health check on our database).

- Support latest AspNetCoreRateLimit lib

The latest version of the `AspNetCoreRateLimit` library supports Redis integration out of the box; this drops our custom Redis counter store and switches to the bundled one.

Minor updates to the `Startup` to support later versions of `SwaggerUI` and `DotEnv`.

- Update Docker to use .NET Core 5

Updates the Docker file base images and the GitHub actions to use v5.


GetIntoTeachingApi/RateLimiting/ApiClientResolveContributor.cs
```diff
@@ -8,27 +9,25 @@ namespace GetIntoTeachingApi.RateLimiting
...
 
-            if (_contextAccessor.HttpContext.Request.Headers.TryGetValue(_clientIdHeader, out var value))
+            if (httpContext.Request.Headers.TryGetValue(_clientIdHeader, out var value))
             {
                 var apiKey = value.ToString().Replace("Bearer ", string.Empty);
                 clientId = _clientManager.GetClient(apiKey)?.ApiKeyPrefix;
             }
 
-            return clientId;
+            return Task.FromResult(clientId);
```
jkempster34, line 30: Has this been updated to async because the update to the library expects it to be?

The `ApiClientResolveContributor` actually does I/O stuff when reading the file. Could that class be updated to async, then no need to use `Task.FromResult`?

Like the highlighted bits here:

image

GetIntoTeachingApi/Models/BaseModel.cs
```diff
@@ -188,7 +188,8 @@ private void NullifyInvalidFieldAttributes(IValidatorFactory validatorFactory)
                 return;
             }
 
-            var result = validator.Validate(this);
+            var context = new ValidationContext<object>(this);
```
jkempster34, line 191: Could this `object` be a `BaseModel`?

GetIntoTeachingApi/RateLimiting/ApiClientResolveContributor.cs
ethax-ross: Yeah the interface was updated to async; there's no benefit to making the `ClientManager.GetClient` async; its a singleton that performs the IO once on boot then we're just reading from memory in that method. In your example you would be hitting the disk on every request to load all the clients into memory and then finding the one that matches which would be more expensive.

GetIntoTeachingApi/Models/BaseModel.cs
ethax-ross: Yep good catch I'll update that and squash it into the relevant commit

GetIntoTeachingApi/RateLimiting/ApiClientResolveContributor.cs
jkempster34: Ah, I see. I just saw `Task.FromResult` and wondered why

jkempster34 (approved):
LGTM 🚴‍♂️

PR #694, opened 2021-07-02
Run dev/test deployments in parallel

ethax-ross:
[Trello-1685](https://trello.com/c/WXcNsFl0/1685-run-the-dev-test-deployments-in-parallel)

At the moment we deploy to the dev environment, then the test environment; it will save about 2.5 minutes off the 10 min deploy time if we deploy both in parallel.

I can't think of any downside to this? other than we get two failures on a bad deploy, but any failure is pretty rare now so probably not a big deal?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -195,7 +195,6 @@ jobs:
            
   qa:
     name: Quality Assurance Deployment
-    needs: [ Development ]
     runs-on: ubuntu-latest
     steps:
        - name: Check out the repo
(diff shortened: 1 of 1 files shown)


sfawcett123 (changes requested):
QA Needs the following settings added, or it will fire at the same time as the build, but there will be no container built.
It needs to be after build, and only on master

    needs: build
    if: github.ref == 'refs/heads/master'

sfawcett123 (approved):

PR #703, opened 2021-07-09
Draft - Explore saving related models in CRM service

jkempster34:
[Commit 1](https://github.com/DFE-Digital/get-into-teaching-api/pull/703/commits/2512f09163d0cc5e27e2f229645a06407dd4bd4c) - `SaveRelated` method that takes a model, related model, and foreign key.

[Commit 2](https://github.com/DFE-Digital/get-into-teaching-api/pull/703/commits/1b2f8d0101a0df078c9ca0cf5ca22edd93ab628e) - Adds a `ForeignKeyAttribute` and loops through related models, saving them before assigning their id to the parent model's foreign key property.

[Commit 3](https://github.com/DFE-Digital/get-into-teaching-api/pull/703/commits/18d508f1738f33bf8fe18ce1e7af0f02473d7bcc) - Make method recursive. Walk through related models, passing parent and child model to method, updating foreign key (Don't know if there would be need for this, do any related models have related models of their own?)

This only works for the one to many relationship of TeachingEvent and Building, I don't think it would be hard to update it to work with the Candidate model and its relationships


jkempster34:
Closing because we have decided to keep things simple and force the caller to create relationships instead

PR #706, opened 2021-07-12
Organise models into namespaces

jkempster34:
There a quite a few models.
- Tidy the `core` models (linked to Dynamics entities) under `Models.Crm`
- Tidy the request/response models under the name of the service they relate to e.g. `Models.GetIntoTeaching`
- Mirror the above folder structure in the tests and validators.
- Also remove all unused `using` statements with help from Visual Studio

Does this look right? Any need moving? :point_down:
image


ethax-ross (commented):
This is looking good, few tweaks I would suggest are:

Controllers/GetIntoTeaching/MailingListController
Models/SchoolsExperience/ClassroomExperienceNote

I would also move the validators into the individual namespaces so:

Models/FindApply/Validators
Models/Crm/Validators
... etc ... 

That way we can consistently keep a singular namespace at the top-level of each subdirectory.

Have you tried re-generating the Api Client locally off the back of these changes to see if it has an impact on the auto-generated code?

jkempster34:
Did this in a fresh PR

PR #708, opened 2021-07-14
Remove relationship entity mapping from `BaseModel`

jkempster34:
- Remove relationship entity mapping from `BaseModel`

We no longer use the MapRelationshipAttributesToEntity method of the BaseModel. This broke after the recent update to the CdsServiceClient, where we now are unable to deep insert entities (create related entities in a single request). We now persist entities independently instead.

### DIFF
--- a/GetIntoTeachingApi/Models/BaseModel.cs
+++ b/GetIntoTeachingApi/Models/BaseModel.cs
@@ -82,7 +82,6 @@ namespace GetIntoTeachingApi.Models
 
             var entity = crm.MappableEntity(LogicalName(GetType()), Id, context);
             MapFieldAttributesToEntity(entity);
-            MapRelationshipAttributesToEntity(entity, crm, context);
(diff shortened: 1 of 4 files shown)


GetIntoTeachingApi/Models/BaseModel.cs
```diff
@@ -268,32 +262,6 @@ private void MapFieldAttributesToEntity(Entity entity)
             }
         }
 
-        private void MapRelationshipAttributesToEntity(Entity source, ICrmService crm, OrganizationServiceContext context)
```
ethax-ross, line 271: Might be worth putting a note in the readme to say that we only map relationships _from_ the CRM to our models and not vice-versa https://github.com/DFE-Digital/get-into-teaching-api#adding-relationships

ethax-ross (approved):

ethax-ross (approved):

PR #709, opened 2021-07-15
Map related model validation errors to request model

jkempster34:
Validation errors on related models will appear in the format `RelatedModel.PropertyName`. It will be easier for clients to handle errors if they are mapped back to the request model property.

- Add an extension to `ValidationResult`. Intercept validation errors and map keys that are duplicated in any related model to the request model. Validation errors will now have the key `PropertyName`.
- Apply the interceptor to any request model validators with a hidden `Candidate` model. Update Swagger controller descriptions to remove warnings about hidden errors.
- Update validation error mapping advise in README

### DIFF
--- a/GetIntoTeachingApi/Utils/ValidationResultExtensions.cs
+++ b/GetIntoTeachingApi/Utils/ValidationResultExtensions.cs
@@ -1,0 +1,31 @@
+﻿using System.Linq;
+using FluentValidation;
+using FluentValidation.Results;
+
+namespace GetIntoTeachingApi.Utils
+{
+    public static class ValidationResultExtensions
+    {
+        public static ValidationResult SurfaceErrorsOnMatchingProperties(
+            this ValidationResult result, IValidationContext validationContext)
+        {
+            var model = validationContext.InstanceToValidate;
+            result.Errors.ForEach(error => RemapError(error, model));
+            return result;
+        }
+
+        private static void RemapError(ValidationFailure error, object model)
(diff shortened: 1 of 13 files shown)


GetIntoTeachingApi/Utils/ValidationResultExtensions.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Linq;
+using FluentValidation;
+using FluentValidation.Results;
+
+namespace GetIntoTeachingApi.Utils
+{
+    public static class ValidationResultExtensions
+    {
+        public static void MapRelatedModelErrorsToRequestModel(
```
ethax-ross, line 9: This is quite a busy method, can it be simplified? Maybe something like:

```
var model = validationContext.InstanceTovalidate;
result.Errors.ForEach(e => RemapError(e, model);
```

```
public void RemapError(error, model)
{
  var errorPropertyName = error.PropertyName.Split('.').Last();
  var modelHasMatchingProperty = model.GetType().GetProperties().Any(p => p.Name == errorPropertyName);

  if (modelHasMatchingProperty)
  {
    error.PropertyName = errorPropertyName;
  }
}
```

ethax-ross (commented):
Looks along the right track, just left a few minor comments. Do you want to update the other request models with this interceptor as part of the PR too?

ethax-ross (approved):
💯 nice!

PR #713, opened 2021-07-19
Increase HighNotFound alert threshold

ethax-ross:
We're seeing this more frequently from just normal traffic/bots poking. Increasing it to 120/10m to hopefully reduce the number of false positives.


leoapost (approved):
Looks good

PR #714, opened 2021-07-20
Updates to GetIntoTeachingCallbackBooking modelling

ethax-ross:
- Update PhoneCall Channel for GiT callback

We need to differentiate between the TTA callback and the GiT callback, which means providing different channels on the PhoneCall entity.

- Add talking points to callback request

When we create a callback from the GiT website we want to collect the talking points/reason for the phone call from the candidate to give the an adviser an indication of what the call will be about.


jkempster34 (approved):
💯

PR #717, opened 2021-07-21
Add Swagger definitions filter

jkempster34:
Since the update to .NET 5.0, Swashbuckle has being including models that should be hidden in the generated Swagger doc. This has gone unnoticed because the API client hasn't been regenerated in a while.

I think this is because of the update to FluentValidation, now that validators must be generic (but I'm not 100% sure)

To prevent this: 
- add a new SwaggerIgnore attribute that can be used to decorate any types that should not be included in the Swagger docs 
- hook into the generation process to inspect the schema for matching types.

### DIFF
--- a/GetIntoTeachingApi/AppStart/ServiceCollectionExtensions.cs
+++ b/GetIntoTeachingApi/AppStart/ServiceCollectionExtensions.cs
@@ -103,9 +103,9 @@ The GIT API aims to provide:
                 });
 
                 c.OperationFilter<AuthOperationFilter>();
+                c.DocumentFilter<SwaggerDefinitionsFilter>();
                 c.EnableAnnotations();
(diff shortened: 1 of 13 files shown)


ethax-ross (approved):
Looks good. Did you try using a `[JsonIgnore]` attribute on the models, as that works for fields? I also wonder if changing the scope of the classes form `public` to `internal` (which arguably they should be anyway) would achieve the same thing.

jkempster34:
> Looks good. Did you try using a `[JsonIgnore]` attribute on the models, as that works for fields? I also wonder if changing the scope of the classes form `public` to `internal` (which arguably they should be anyway) would achieve the same thing.

I did try `internal` yeah, I'll try JsonIgnore now

PR #719, opened 2021-07-22
Namespace and move models and controllers

jkempster34:
# Trello

1. [namespacing](https://trello.com/c/vWMQwbqe)
2. [renaming `Candidate`, which I think we agreed this would be adequate instead](https://trello.com/c/HyZLtrgE)

# Context
We want to differentiate between the different model/controller types as the app grows in size.

# Changes
- Move and namespace controllers `MailingListController`, `TeachingEventBuildingController` and `TeachingEventController` under `GetIntoTeaching`
- Move and namespace all `core` models under `Crm`
- Move and namespace request models that are client specific under client name
- Move validators into individual namespaces along with models
- Update file structure and namespacing of tests to match the above
- Allow Visual Studio to reorder and remove unused `using` statements

# Guidance
**Models** 👇 
image

**Controllers** 👇 
image


ethax-ross (approved):
Looks great 💯 

I think a final bit we should do at some point is to change the groupings in Swagger to match; I think this is already the case for the TTA and SE endpoints but not for GiT. That will definitely effect the API client though, so it'll need a bit of coordinating to do without downtime.

PR #724, opened 2021-07-29
Check if candidate can re-subcsribe based on adviser status

ethax-ross:
[Trello-1698](https://trello.com/c/yG3hzAuh/1698-development-unblock-get-an-adviser-applicants-who-have-previously-had-an-adviser)

Up until now if a candidate has already been assigned an adviser we prevent them from re-signing up to the TTA service.

It turns out some candidates that have been assigned an adviser will have their `contact` record set to have a 'closed' adviser status. In this scenario, we want to let them re-subscribe for an adviser.

Adds a new `CanSubscribeToTeacherTrainingAdviser` attribute as `AlreadySubscribedToTeacherTrainingAdviser` would not be completely accurate for the new use case. Once the TTA service has been updated this field can be removed.

### DIFF
--- a/GetIntoTeachingApi/Models/TeacherTrainingAdviser/TeacherTrainingAdviserSignUp.cs
+++ b/GetIntoTeachingApi/Models/TeacherTrainingAdviser/TeacherTrainingAdviserSignUp.cs
@@ -10,6 +10,39 @@ namespace GetIntoTeachingApi.Models.TeacherTrainingAdviser
 {
     public class TeacherTrainingAdviserSignUp
     {
+        public enum ResubscribableAdviserStatus
+        {
+            AcceptedIttOffer = 222750000,
+            AcceptedRttTeachingPosition = 222750004,
+            AlreadyHasQts = 222750005,
+            AskedToBeLeftAlone = 222750006,
+            IncorrectlyAllocated = 222750007,
+            NoLongerPursuingTeaching = 222750008,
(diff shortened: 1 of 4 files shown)


GetIntoTeachingApi/Models/TeacherTrainingAdviser/TeacherTrainingAdviserSignUp.cs
```diff
@@ -63,6 +98,22 @@ public TeacherTrainingAdviserSignUp(Candidate candidate)
...
+            {
+                return true;
+            }
+
+            if (candidate.AdviserStatus == null)
+            {
+                return false;
+            }
+
+            return Enum.GetValues(typeof(ResubscribableAdviserStatus))
```
jkempster34, line 113: Wondering if `Enum.IsDefined` would work here? http://www.blackwasp.co.uk/EnumIsDefined.aspx

ethax-ross: Yep good shout; I dismissed that earlier because it wouldn't handle the case when the field is `null` but I didn't get that working with `GetValues` either (for some reason it wouldn't let me cast the `int` of the enum to `int?` of the field 🤷‍♂️) so I broke out the `null` check and stayed with `GetValues`; I'll push up the change

jkempster34 (approved):
🥇

PR #728, opened 2021-07-29
Handle candidates re-signing up with a closed adviser status

ethax-ross:
[Trello-1698](https://trello.com/c/yG3hzAuh/1698-development-unblock-get-an-adviser-applicants-who-have-previously-had-an-adviser)

- Rename AdviserStatus to AdviserStatusId

To match the other option set fields.

- Expose AdviserStatusId on TeacherTrainingAdviserSignUp

We need to expose this so that the client can post it back to us for matchback scenarios (otherwise we would have to query the CRM to retrieve it, which would complicate the logic for when the CRM is offline etc).

- Handle candidates with closed adviser status

When a candidate re-subscribes and has a closed adviser status we want to:

```
- Set the `AssignmentStatusId` to `WaitingToBeAssigned`
- Set the `RegistrationStatusId` to `ReRegistered`
- Update the `StatusIsWaitingToBeAssignedAt` to the current date/time
```

- Post adviser_status_id back to the API

If an candidate has an `adviser_status_id` returned as part of the matchback process we want to post this back to the API as part of the submission. This avoids the API having to re-retrieve the candidate in order to determine the status.

### DIFF
--- a/GetIntoTeachingApi/Models/Crm/Candidate.cs
+++ b/GetIntoTeachingApi/Models/Crm/Candidate.cs
@@ -100,6 +100,11 @@ namespace GetIntoTeachingApi.Models.Crm
             Exchanged = 222750004,
         }
 
+        public enum RegistrationStatus
+        {
+            ReRegistered = 222750000,
+        }
+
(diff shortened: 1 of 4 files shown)


jkempster34 (approved):
LGTM

PR #736, opened 2021-08-04
Set event registration channel for walk-ins

ethax-ross:
> :warning: **pending CRM changes shipping to prod**

When a candidate signs up as a 'walk in' at the event, we need to make sure the `TeachingEventRegistration` is created with the appropriate channel.


jkempster34 (approved):
LGTM 👍

PR #739, opened 2021-08-05
Map AdviserStatusId from Candidate

ethax-ross:
We were not mapping the `AdviserStatusId` to the request model from the `Candidate`, so when used to pre-fill the TTA sign up the field was missing and the behaviour was not correct on the subsequent form submission.


jkempster34 (approved):

ethax-ross:
Github actions got stuck so I had to force-push the same commit; no changes since last approval so just going to merge down.

PR #741, opened 2021-08-06
Bug fix to support empty array in GetCandidates method

ethax-ross:
The Dynamics `QueryExpression` interpreter doesn't support an empty value for an `In` conditional, so we need to return early in this instance and not call the CRM.


jkempster34 (approved):
💯

PR #744, opened 2021-08-09
Fix flakey contract tests

ethax-ross:
Bit of a shot in the dark here, but the tests fail fairly routinely on CI although I can't replicate it locally. As the contract is based on the jobs in progress I'm betting that another test leaves left over state in the job queue that is picked up when the first contract test runs.

To avoid this, we're going to flush the state before each test run (clear the job queue, wait for any in-progress jobs to finish and reset the `TrackedEntities`).


jkempster34 (approved):
Makes sense 👍

PR #748, opened 2021-08-09
Check for zip bombs before extracting postcodes

jkempster34:
[Trello](https://trello.com/c/RrDgOoJI)

SonarCloud is showing that the API is vulnerable to Zip Bomb attacks. By checking the compression ratio, total number of archives, and total archive size, we can assure that the Zip file is not malicious.

- Create a `ZipFileChecker` utility which defaults to recommended limits for the above attributes. It will throw a `BombFoundException` if the limits are exceeded.


GetIntoTeachingApi/Jobs/LocationSyncJob.cs
```diff
@@ -149,13 +149,18 @@ private async Task<string> RetrieveCsv(string ukPostcodeCsvUrl)
 
             try
             {
+                ZipFileChecker.AssureNoBombs(zipPath);
                 ZipFile.ExtractToDirectory(zipPath, csvPath);
-                _logger.LogInformation($"LocationSyncJob - CSV Extracted");
+                _logger.LogInformation("LocationSyncJob - CSV Extracted");
+            }
+            catch (BombFoundException bombFoundException)
```
ethax-ross, line 156: I would let this bubble up so we get a notification in Sentry

GetIntoTeachingApi/Utils/ZipFileChecker.cs
```diff
@@ -0,0 +1,67 @@
...
+                        totalArchiveSize += numberOfBytesRead;
+                        double compressionRatio = totalEntrySize / (double)entry.CompressedLength;
+
+                        if (compressionRatio > MaxCompressionRatio)
+                        {
+                            throw new BombFoundException(
+                                $"Compression ratio of ${compressionRatio} exceeds the maximum allowed ${MaxCompressionRatio}");
+                        }
+                    }
+                    while (numberOfBytesRead > 0);
```
ethax-ross, line 46: It looks like if we got the first entry here that was > MaxTotalArchiveSize it would read the entire entry before raising the exception, which defeats the purpose as it could use up all the disc space; I think we need to fail faster - so check the total archive size within the do/while loop.

ethax-ross (commented):
Looks good, just a couple comments -- it would be good to be able to test this though; I wonder if we could include a zip bomb as a fixture and perhaps make the limits configurable so we can make them small for a test run?

GetIntoTeachingApiTests/Utils/ZipFileCheckerTests.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            Action check = () => checker.AssureNoBombs(suspiciousZipFilePath);
+
+            check.Should()
+                .Throw<BombFoundException>()
+                .WithMessage("Found 2 entries which exceeds the maximum allowed (1)");
+        }
+
+        [Fact]
+        public void AssureNoBombs_WhenArchiveSizeIsSuspicious_ThrowsBombFoundException()
```
ethax-ross, line 37: May be worth adding a happy path test case here as well?

GetIntoTeachingApiTests/Utils/ZipFileCheckerTests.cs
```diff
@@ -0,0 +1,50 @@
...
+            check.Should()
+                .Throw<BombFoundException>()
+                .WithMessage("Compression ratio of 25 exceeds the maximum allowed (10)");
+        }
+
+        [Fact]
+        public void AssureNoBombs_WhenNumberOfEntriesIsSuspicious_ThrowsBombFoundException()
+        {
+            const string suspiciousZipFilePath = "./Fixtures/zip-bomb-fixtures/contains-two-files.zip";
+            var checker = new ZipFileChecker(maxNumberOfEntries: 1);
```
ethax-ross, line 27: Named arguments 🤤

GetIntoTeachingApi/Jobs/LocationSyncJob.cs
```diff
@@ -145,17 +145,24 @@ private async Task<string> RetrieveCsv(string ukPostcodeCsvUrl)
...
 
             try
             {
+                var checker = new ZipFileChecker();
+                checker.AssureNoBombs(zipPath);
                 ZipFile.ExtractToDirectory(zipPath, csvPath);
-                _logger.LogInformation($"LocationSyncJob - CSV Extracted");
+                _logger.LogInformation("LocationSyncJob - CSV Extracted");
+            }
+            catch (BombFoundException exception)
```
ethax-ross, line 157: I would drop the catch and log, the fact that Sentry will flag it is enough

ethax-ross (approved):
Looks great 💯

PR #752, opened 2021-08-11
Format logging in JSON with Serilog

jkempster34:
[Trello](https://trello.com/c/rIFDcBnt)

ASP.NET log messages aren't interpreted properly by Logit.
- Set Serilog as the logging provider, so that all ILoggers now use it and log messages in JSON.
- Remove default logger configuration.

### DIFF
--- a/GetIntoTeachingApi/Program.cs
+++ b/GetIntoTeachingApi/Program.cs
@@ -2,6 +2,8 @@ using System.Threading.Tasks;
 using GetIntoTeachingApi.AppStart;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Hosting;
+using Serilog;
+using Serilog.Events;
 
 namespace GetIntoTeachingApi
 {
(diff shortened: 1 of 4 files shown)


GetIntoTeachingApi/Program.cs
```diff
@@ -22,6 +25,12 @@ public static async Task Main(string[] args)
                     webBuilder.UseSentry();
                     webBuilder.UseKestrel(opts => opts.AddServerHeader = false);
                     webBuilder.UseStartup<Startup>();
-                });
+                })
+            .UseSerilog(new LoggerConfiguration()
+                .MinimumLevel.Information()
+                .MinimumLevel.Override("Microsoft", LogEventLevel.Warning)
+                .WriteTo.Console()
+                .WriteTo.TCPSink("") // Logstash TCP URI needed here
```
ethax-ross, line 33: I don't think this is needed? I believe `WriteTo.Console` will go to the STDOUT of the Docker container, which is what's piped through to our logstash

ethax-ross (commented):
Looks good just one comment - didn't realise Serilog adheres to the `ILogger` interface, makes it a whole bunch easier!

GetIntoTeachingApi/Program.cs
jkempster34: Ahh, okay!

ethax-ross (approved):

PR #753, opened 2021-08-12
Allow multiple event types in event search

jkempster34:
https://trello.com/c/qbuoFw8L

We may want to search by multiple event types to prevent filtering in the client. Update the type ID parameter of the teaching event search to an array of event types.

Not sure how we would want to handle multiple event types in Prometheus. Not sure if it really matters, but do we need a "don't track" param so that we don't export certain searches (internal searches) to Prometheus.

Also, not sure what the best way to clean up the two PickListItem validators is...

### DIFF
--- a/GetIntoTeachingApi/Controllers/GetIntoTeaching/TeachingEventsController.cs
+++ b/GetIntoTeachingApi/Controllers/GetIntoTeaching/TeachingEventsController.cs
@@ -78,13 +78,19 @@ namespace GetIntoTeachingApi.Controllers.GetIntoTeaching
 
             var teachingEvents = await _store.SearchTeachingEventsAsync(request);
 
+            // TODO: Fix. Prometheus can only be passed two labels currently. Don't know how we would want to handle multiple type IDs.
+            List<string> labels = new ();
+            var typeIds = request.TypeIds?.Select(id => id.ToString());
+            labels.Add(typeIds == null ? string.Empty : typeIds.First());
+            labels.Add(request.Radius.ToString());
+
             _metrics.TeachingEventSearchResults
-                .WithLabels(request.TypeId.ToString(), request.Radius.ToString())
+                .WithLabels(labels.ToArray())
                 .Observe(teachingEvents.Count());
 
             var inPesonTeachingEvents = teachingEvents.Where(e => e.IsInPerson);
             _metrics.InPersonTeachingEventResults
-                .WithLabels(request.TypeId.ToString(), request.Radius.ToString())
+                .WithLabels(labels.ToArray())
                 .Observe(inPesonTeachingEvents.Count());
 
             return Ok(GroupTeachingEventsByType(teachingEvents, quantityPerType));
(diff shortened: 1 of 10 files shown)


GetIntoTeachingApi/Controllers/GetIntoTeaching/TeachingEventsController.cs
```diff
@@ -78,13 +78,19 @@ public class TeachingEventsController : ControllerBase
 
             var teachingEvents = await _store.SearchTeachingEventsAsync(request);
 
+            // TODO: Fix. Prometheus can only be passed two labels currently. Don't know how we would want to handle multiple type IDs.
```
ethax-ross, line 81: You can pass more than two but you need to define them explicitly on the metric here:

https://github.com/DFE-Digital/get-into-teaching-api/blob/master/GetIntoTeachingApi/Services/MetricService.cs#L23

I would rename the `type_id` metric to `type_ids` and pass it a comma separated list in this case (it would be more flexible if we had a label for each type id but it uses a bunch more space and we don't need that for what we're reporting on here)

GetIntoTeachingApi/Models/GetIntoTeaching/TeachingEventSearchRequest.cs
```diff
@@ -11,7 +11,7 @@ public class TeachingEventSearchRequest : ICloneable
         [SwaggerSchema("Set to filter results to a radius (in miles) around the postcode.")]
         public int? Radius { get; set; }
         [SwaggerSchema("Set to filter results to a type of teaching event. Must match an `typeId` of the `TeachingEvent` schema.")]
-        public int? TypeId { get; set; }
```
ethax-ross, line 14: This will need to stay in for backwards-compatibility initially; we can remove it once we've updated the clients to use `TypeIds`

GetIntoTeachingApi/Models/GetIntoTeaching/Validators/TeachingEventSearchRequestValidator.cs
```diff
@@ -13,9 +13,9 @@ public TeachingEventSearchRequestValidator(IStore store)
                 .MaximumLength(40)
                 .Matches(Location.OutwardOrFullPostcodeRegex)
                 .Unless(request => request.Postcode == null && request.Radius == null);
-            RuleFor(request => request.TypeId)
-                .SetValidator(new PickListItemIdValidator<TeachingEventSearchRequest>("msevtmgt_event", "dfe_event_type", store))
-                .Unless(request => request.TypeId == null);
+            RuleFor(request => request.TypeIds)
```
ethax-ross, line 16: `FluentValidation` has a `RuleForEach` that will let you use the current validator that works on a single pick list item id, for example:

```
RuleForEach(r => r.TypeIds)
  .SetValidator(new PickListItemIdValidator<TeachingEventSearchRequest>("msevtmgt_event", "dfe_event_type", store));
```

PR #754, opened 2021-08-17
Add contract test for re-subscribing to a TTA

ethax-ross:
When a candidate is in a 'closed' state (their `adviserStatusId` is one that is 'closed') we allow them to re-subscribe to the TTA service.

Add contract test for a candidate re-subscribing for a TTA, capturing the requests made to the CRM.


peteryates (approved):
Looks good.

PR #759, opened 2021-08-20
Update FluentValidation dependency

ethax-ross:
- Update FluentValidation

- Address deprecation warnings for FluentValidation

The `validator.ShouldHaveValidationErrorFor(attr, value)` has been deprecated in favour of using `TestValidate`. This updates all references, consolidates some of the separate tests into single tests to make them easier to read and removes some of the assertions for the inverse of rules (checking `null` is allowed, for example) as these don't add much value and bloat the test  suite quite a bit.


peteryates (approved):
Nice - looks much cleaner

PR #768, opened 2021-08-23
Update FreeMapToolsUrl with new CSV path

ethax-ross:
The FreeMapTools service we use to geocode the bulk of our postcodes have changed the path to their CSV file. This updates it; the format appears unchanged.


peteryates (approved):
Link works, looks good to me

PR #773, opened 2021-08-31
Allow multiple event types in event search

jkempster34:
[Trello](https://trello.com/c/qbuoFw8L)

Sometimes we want to search by multiple event types so that we don't have to filter events in the client. 

For now we are keeping the `TypeId` property for backwards compatibility. It will be removed when the clients have been updated. 

- Add a `TypeIds` property that accepts an array of type IDs.
- Update any use of `TypeId` to `TypeIds` throughout the project, and temporarily set `TypeIds` to the value of `TypeId` if it is present.
- Update the `TeachingEventSearchRequestValidator` to a non-nullable `int`. This only broke one test, so I think it should be okay.

### DIFF
--- a/GetIntoTeachingApi/Controllers/GetIntoTeaching/TeachingEventsController.cs
+++ b/GetIntoTeachingApi/Controllers/GetIntoTeaching/TeachingEventsController.cs
@@ -78,13 +78,15 @@ namespace GetIntoTeachingApi.Controllers.GetIntoTeaching
 
             var teachingEvents = await _store.SearchTeachingEventsAsync(request);
 
+            var typeIds = request.TypeIds == null ? string.Empty : string.Join(",", request.TypeIds);
+
             _metrics.TeachingEventSearchResults
-                .WithLabels(request.TypeId.ToString(), request.Radius.ToString())
+                .WithLabels(typeIds, request.Radius.ToString())
                 .Observe(teachingEvents.Count());
 
             var inPesonTeachingEvents = teachingEvents.Where(e => e.IsInPerson);
             _metrics.InPersonTeachingEventResults
-                .WithLabels(request.TypeId.ToString(), request.Radius.ToString())
+                .WithLabels(typeIds, request.Radius.ToString())
                 .Observe(inPesonTeachingEvents.Count());
 
             return Ok(GroupTeachingEventsByType(teachingEvents, quantityPerType));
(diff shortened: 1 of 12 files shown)


GetIntoTeachingApi/Models/Crm/Validators/CandidateValidator.cs
```diff
@@ -59,6 +59,7 @@ public CandidateValidator(IStore store, IDateTimeProvider dateTime)
                 .SetValidator(new PickListItemIdValidator<Candidate>("contact", "dfe_ittyear", store))
                 .Unless(candidate => candidate.InitialTeacherTrainingYearId == null);
             RuleFor(candidate => candidate.ChannelId)
+                .NotNull()
```
ethax-ross, line 62: I think there's a mistake on line 64 that if we correct should mean it doesn't need this `NotNull` -- it should be `Unless(c => c.Id == null)` but if we want it to be required we can just remove the `Unless` clause and have only the `SetValidator` which will enforce it to be one of the valid values and not `null`.

GetIntoTeachingApi/Services/MetricService.cs
```diff
@@ -20,12 +20,12 @@ public class MetricService : IMetricService
         private static readonly Histogram _teachingEventSearchResults = Metrics
             .CreateHistogram("api_teaching_event_search_results_count", "Histogram of teaching event search results.", new HistogramConfiguration
             {
-                LabelNames = new[] { "type_id", "radius" },
+                LabelNames = new[] { "type_ids", "radius" },
```
ethax-ross, line 23: We'll need to update some of the Grafana panels once this changes (probably just on the search dashboard)

GetIntoTeachingApi/Controllers/GetIntoTeaching/TeachingEventsController.cs
```diff
@@ -78,9 +78,20 @@ public class TeachingEventsController : ControllerBase
 
             var teachingEvents = await _store.SearchTeachingEventsAsync(request);
 
-            _metrics.TeachingEventSearchResults
-                .WithLabels(request.TypeId.ToString(), request.Radius.ToString())
-                .Observe(teachingEvents.Count());
+            // Temporary condition, to be removed when clients are updated and TypeId is removed
+            if (request.TypeIds != null && request.TypeId == null)
```
ethax-ross, line 82: I think this and the updates to the store could be simplified by instead updating the request model so that you override the `TypeIds` getter to either return its value (if set) or `new int[] { TypeId }` if null and `TypeId` is set. That way you don't need to have two sets of conditionals in the controller/store to account for the different methods and you can just use `request.TypeIds` everywhere.

GetIntoTeachingApi/Models/Crm/Validators/CandidateValidator.cs
jkempster34: This is to satisfy this test https://github.com/DFE-Digital/get-into-teaching-api/blob/bab56c6b81ecf94ef8c835d7c86d22bf437849b3/GetIntoTeachingApiTests/Models/Crm/Validators/CandidateValidatorTests.cs#L495

ethax-ross (approved):
Looks good, I would double check the event search from the front-end in dev before shipping this out to prod but seems fine 👍

jkempster34:
> Looks good, I would double check the event search from the front-end in dev before shipping this out to prod but seems fine 👍

Will do 🙏

PR #774, opened 2021-09-02
Increase GoogleApiErrors alert threshold

ethax-ross:
We are seeing the occasional `ZERO_RESULTS` response from Google when someone inputs an unknown postcode that is in a valid format. Instead, we want this to alert us if there is a flurry of bad requests to Google (most likely caused by an event that is given an invalid postcode or our Google API key being revoked).


jkempster34 (approved):
👌

PR #776, opened 2021-09-03
Extend comma separated attribute so that it can handle complex types

jkempster34:
[Trello](https://trello.com/c/qbuoFw8L)

We want to search events by multiple TypeIds. To do this we have added a new collection type parameter. Unfortunately, ASP.NET expects the request in the following format:

```
?TypeIds=101&TypeIds=202
```

whereas the Swagger Codegen API client library produces incompatible queries that look like this:

```
?TypeIds=101,202
``` 

There is already a `CommaSeparatedAttribute` which handles this, but it only works for simple types because the `CommaSeparatedQueryStringConvention` passes the name of the parameter itself to the `CommaSeparatedQueryStringValueProvider`.

Extend the attribute so that we can specify which properties of complex types should be comma separated.

### DIFF
--- a/GetIntoTeachingApi/Attributes/CommaSeparatedQueryStringAttribute.cs
+++ b/GetIntoTeachingApi/Attributes/CommaSeparatedQueryStringAttribute.cs
@@ -19,9 +19,9 @@ namespace GetIntoTeachingApi.Attributes
             _factory = new CommaSeparatedQueryStringValueProviderFactory(separator);
         }
 
-        public CommaSeparatedQueryStringAttribute(string key, string separator)
+        public CommaSeparatedQueryStringAttribute(string[] keys, string separator)
         {
-            _factory = new CommaSeparatedQueryStringValueProviderFactory(key, separator);
+            _factory = new CommaSeparatedQueryStringValueProviderFactory(keys, separator);
         }
 
         public void OnResourceExecuted(ResourceExecutedContext context)
--- a/GetIntoTeachingApi/Utils/CommaSeparatedQueryStringConvention.cs
+++ b/GetIntoTeachingApi/Utils/CommaSeparatedQueryStringConvention.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System.Collections.Generic;
+using System.Linq;
 using GetIntoTeachingApi.Attributes;
 using Microsoft.AspNetCore.Mvc.ApplicationModels;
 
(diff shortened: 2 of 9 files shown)


GetIntoTeachingApi/Attributes/CommaSeparatedQueryStringAttribute.cs
```diff
@@ -19,9 +19,9 @@ public CommaSeparatedQueryStringAttribute(string separator)
             _factory = new CommaSeparatedQueryStringValueProviderFactory(separator);
         }
 
-        public CommaSeparatedQueryStringAttribute(string key, string separator)
+        public CommaSeparatedQueryStringAttribute(string[] keys, string separator)
```
ethax-ross, line 22: `IEnumerable<string> keys` would be more flexible here

GetIntoTeachingApi/Utils/CommaSeparatedQueryStringConvention.cs
```diff
@@ -12,9 +13,23 @@ public void Apply(ActionModel action)
...
+                        AddToActionFilterAsCommaSeparated(parameter, new string[] { parameter.ParameterName });
+                    }
                 }
             }
-        }
+        }
+
+        private static void AddToActionFilterAsCommaSeparated(ParameterModel parameter, string[] keys)
+        {
+            parameter.Action.Filters.Add(new CommaSeparatedQueryStringAttribute(keys, ","));
```
ethax-ross, line 32: Instead of bubbling `keys` all the way down to `CommaSeparatedQueryStringAttribute` would it be clearer/fewer changes to call it multiple times? something like:

```
var attributeNames = parameter.Attributes.OfType<CommaSeparatedAttribute>().First().AttributeNames;
if (attributeNames == null) {
  attributeNames = new string[] { parameter.ParameterName };
}

attributeNames.ForEach(a => parameter.Action.Filters.Add(new CommpaSeparatedQueryStringAttribute(a, ",");
```

ethax-ross (approved):
Looks good, couple suggestions but I don't feel all that strongly about them.

GetIntoTeachingApi/Utils/CommaSeparatedQueryStringConvention.cs
jkempster34: The `CommaSeparatedQueryStringValueProvider` gets called once per request with a collection of the query parameters so I don't think this will work, it needs access to a list of allowed comma separated keys to check against.

PR #783, opened 2021-09-07
Undo remapping of QT events to TTT

ethax-ross:
The front-end has been updated to natively support QT events by displaying them as though they are TTT events. We should now be fine to remove the remapping that occurs in the API.

This will need to be merged down and checked on dev/test before shipping to prod.


jkempster34 (approved):
👍

PR #791, opened 2021-09-15
Make first/last name optional on Candidate

ethax-ross:
[Trello-1930](https://trello.com/c/kcWhKB9W/1930-support-syncing-apply-only-candidates-in-the-api)

The candidates that come through from the Apply API will not always have a `FirstName` or `LastName`. We need to be able to accept records where a name has not been set, but still enforce first/last name on sign ups. As the request models already have validation in place to ensure first/last name is present, we can simply remove the `NotEmpty` check from the base `Candidate` model.

The existing validation on `Candidate` that was checking for required fields included some that are not required but were failing because an empty string is not valid; this is covered elsewhere so the test has been reduced to only include the required `Email` field.


jkempster34 (approved):
👌

PR #792, opened 2021-09-15
Sync new Apply candidates to the CRM

ethax-ross:
[Trello-1930](https://trello.com/c/kcWhKB9W/1930-support-syncing-apply-only-candidates-in-the-api)

- Sync new Apply candidates to the CRM

Now that first/last name is not required in the base `Candidate` model we can map new Apply API candidates and persist them back to the CRM.

- Set channel for new Apply candidates

When we create a candidate through the Apply API integration we need to ensure the correct channel is set so that it can be attributed to that source.

- Fallback to matching on email only

We are now accepting candidates from Apply that only have an email address and no first/last name, however we need to be able to match back on these candidate records. In order to do this, we first attempt a 'full' match back on first/last/email and, if that fails, we perform a second match back lookup using only the email address.


jkempster34 (approved):
All looks good to me! 💯

PR #794, opened 2021-09-16
Use FluentAssertions.Json in contract test assertion

ethax-ross:
- Use FluentAssertions.Json in contract test assertion

The contract tests are flakey and can fail on occasion in CI. I'm not sure how this is happening yet, but in an effort to debug it (and maybe even resolve it) we're switching to use `FluentAssertions.Json` in order to check equivalency of of the request JSON with that of our snapshot JSON. If it fails again this should hopefully give us a more detailed diff of where it's is matched.

- Sort entities and attributes prior to comparison

Looking at the diff when the contract test fails in CI it appears to be because the ordering of the entity attributes has changed. As the attributes are not a collection the ordering is probably undefined and we can't sort it prior to serialising. Instead, we deserialise the snapshot and then sort it manually before performing the comparison.

### DIFF
--- a/GetIntoTeachingApiTests/Contracts/TeacherTrainingAdviserTests.cs
+++ b/GetIntoTeachingApiTests/Contracts/TeacherTrainingAdviserTests.cs
@@ -13,6 +13,7 @@ using GetIntoTeachingApi.Models;
 using GetIntoTeachingApi.Models.Crm;
 using GetIntoTeachingApiTests.Helpers;
 using Hangfire;
+using MoreLinq;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using Xunit;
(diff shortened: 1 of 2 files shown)


GetIntoTeachingApiTests/Contracts/TeacherTrainingAdviserTests.cs
```diff
@@ -59,7 +61,21 @@ public async void Contract(string scenario)
...
+            var request = SortEntities(JArray.Parse(requestJson));
+            var snapshot = SortEntities(JArray.Parse(File.ReadAllText(outputFile)));
+
+            request.Should().HaveCount(snapshot.Count());
+            request.Should().BeEquivalentTo(snapshot);
+        }
+
+        private static JArray SortEntities(JArray entities)
+        {
+            for(var i = 0; i < entities.Count; i++)
```
jkempster34, line 73: Wondering if this for loop is left over from something else you tried and could be a foreach?

jkempster34 (dismissed):
Nice 🔥. Glad I didn't have to attempt to fix it

GetIntoTeachingApiTests/Contracts/TeacherTrainingAdviserTests.cs
ethax-ross: Ah yeah that could be cleaned up a bit, I tried a whole bunch of things to make it recursively sort `JArray` and `JObject` but it made it way less readable overall and we don't seem to nest more than one level anyway. I'll fixup a commit.

jkempster34 (approved):

PR #795, opened 2021-09-20
Add Prometheus alert for any CSP violations that include `Google`

jkempster34:
We are going to remove a large CSP whitelist of Google domains in the app because the headers are too large. Therefore, we will want to be alerted when there are any CSP violations that include `Google` so that the list can be built back up.


ethax-ross (approved):
I'm not certain but I think you may just want `sum()` and not `sum by()` - `sum()` should return a single data point result, whereas `sum by()` would return a data point per `blocked_uri`. I don't know how that then gets compared to `> 0` - it may just work, but I haven't tried it.

PR #796, opened 2021-09-21
Ignore legacy paths in HighNotFound alert

ethax-ross:
The `HighNotFound` alert goes off quite often due to either bots poking about or inbound links to legacy website URLs. This attempts to strip out some of the noise around requests that 404 but are not a sign of any underlying issue.


jkempster34 (approved):
Sorry, this got buried under the CSP alerts. LGTM 👍

PR #799, opened 2021-09-22
Swallow errors when there are concurrent requests to add a postcode

jkempster34:
When a user searches for a postcode that isn't in the cache, the API will look it up via Google Geocoding then persist it.

If the API gets multiple requests in quick succession, the requests will be processed concurrently, resulting in a duplicate key exception as the API tries to persist the same postcode multiple times (the key for the Locations table is the postcode).

- Swallow the error in this very specific scenario.

### DIFF
--- a/GetIntoTeachingApi/Services/Store.cs
+++ b/GetIntoTeachingApi/Services/Store.cs
@@ -342,7 +342,14 @@ namespace GetIntoTeachingApi.Services
         {
             var location = new Location(postcode, coordinate, Location.SourceType.Google);
             await _dbContext.Locations.AddAsync(location);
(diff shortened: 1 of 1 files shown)


ethax-ross (approved):
I think this is fine; the only other options I can think of are to use the upsert lib for EF Core which would support this or to write it as a native query that uses Postgres `ON CONFLICT DO NOTHING`, similar to what I did for the bulk import:

https://github.com/DFE-Digital/get-into-teaching-api/blob/master/GetIntoTeachingApi/Jobs/LocationSyncJob.cs#L123-L124

You're right that there's no easy way to test this though; we would need to be able to mock the `_dbContext` read operation to always return empty but its just not possible with Moq.

PR #800, opened 2021-09-23
Add google csp table grafana

jkempster34:
- Fix existing Grafana CSP panels - There are two repeating rows of Grafana panels for CSP violations for the GiT app and TTA app. At some point these have broken, possibly due to an update.
image

- Add a Grafana table of Google CSP violations - We want to build up a CSP whitelist of Google domains in the GiT app. Adding a table showing the last 24 hours of violations so that we can have an alert posted to Slack.
image

- Add daily alert if /there are Google CSP violations - This alert will fire at 10am daily if there are any Google violations in the last 24 hours, and will contain a link to a table displaying them.

- Add Prometheus config - We can declare rules for certain alerts in Prometheus groups. Ensure that the Google CSP alert only fires once rather than continuously during the time specified in the alert rule.

### DIFF
--- a/monitoring/prometheus/alert.rules
+++ b/monitoring/prometheus/alert.rules
@@ -75,6 +75,14 @@ groups:
           description: Alerts when any of the app instances exceed 70% memory utilisation.
           runbook: https://dfedigital.atlassian.net/wiki/spaces/GGIT/pages/2156036281/App+TTA+Runbook#HighMemory-MEDIUM
           dashboard: https://grafana-prod-get-into-teaching.london.cloudapps.digital/d/0PRnzc2Mk/get-into-teaching-apps?viewPanel=33&orgId=1
+      - alert: GoogleCspViolation
+        expr: 'sum(increase(app_csp_violations_total{blocked_uri=~"google"}[24h])) and ON() hour() == 11'
+        labels:
+          severity: low
+        annotations:
+          summary: Alerts when there have been Google CSP violations in production in the last 24 hours.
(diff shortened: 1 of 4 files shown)


monitoring/prometheus/alert.rules
```diff
@@ -75,6 +75,14 @@ groups:
           description: Alerts when any of the app instances exceed 70% memory utilisation.
           runbook: https://dfedigital.atlassian.net/wiki/spaces/GGIT/pages/2156036281/App+TTA+Runbook#HighMemory-MEDIUM
           dashboard: https://grafana-prod-get-into-teaching.london.cloudapps.digital/d/0PRnzc2Mk/get-into-teaching-apps?viewPanel=33&orgId=1
+      - alert: GoogleCspViolation
+        expr: 'sum(increase(app_csp_violations_total{blocked_uri=~"google"}[24h])) and hour() == 11 and minute() == 0'
```
jkempster34, line 79: Just realised this might fire continuously for the minute. Any other way to do this?

ethax-ross: Yeah thats going to be a problem, also I'm not sure how often alert manager evaluates it may not be every minute so may not fire at all.

I would try something like `... and ON() hour() == 11` in the expression to make it only fire sometime during 11-12 and then look into Alert Manager routes; if you make this alert bubble into a different route (they are all using the default route atm) you should be able to configure it with a `repeat_interval` to be > 1 hour, that way if it fires at some time during 11-12 it can't then fire until the following days 11-12 time period.

https://prometheus.io/docs/alerting/latest/configuration/#route

ethax-ross (approved):
LGTM

PR #803, opened 2021-09-29
Stricter partial URL validation

jkempster34:
- Add stricter validation to teaching event readable ID (partial URL) - Currently there is only minimal validation on the event readable ID (partial URL), so we have seen partial URLs with non-alphanumeric characters such as brackets, colons, and question marks. Update the partial URL validation so that only alphanumeric, hyphens, and underscores are allowed.

- Use FluentValidation test helpers - Minor clean-up to ensure unique readable ID validations are coming through on the correct property.

### DIFF
--- a/GetIntoTeachingApiTests/Models/GetIntoTeaching/Validators/TeachingEventUpsertOperationValidatorTests.cs
+++ b/GetIntoTeachingApiTests/Models/GetIntoTeaching/Validators/TeachingEventUpsertOperationValidatorTests.cs
@@ -69,7 +69,11 @@ namespace GetIntoTeachingApiTests.Models.GetIntoTeaching.Validators
         [InlineData("event?name", true)]
         [InlineData("event@name", true)]
         [InlineData("event:name", true)]
-        public void Validate_ReadableId_AllowsOnlyAlphanumericHyphensUnderscores(string readableId, bool hasError)
+        [InlineData("eventname-", true)]
+        [InlineData("-eventname", true)]
+        [InlineData("eventname_", true)]
+        [InlineData("_eventname", true)]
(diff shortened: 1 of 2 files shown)


GetIntoTeachingApiTests/Models/GetIntoTeaching/Validators/TeachingEventUpsertOperationValidatorTests.cs
```diff
@@ -40,22 +41,47 @@ public void Validate_WhenExistingEventWithReadableIdHasSameTeachingEventId_HasNo
...
+        [Theory]
+        [InlineData("eventname", false)]
+        [InlineData("eventname1", false)]
+        [InlineData("event-name", false)]
+        [InlineData("event_name", false)]
+        [InlineData("event name", true)]
+        [InlineData("event?name", true)]
+        [InlineData("event@name", true)]
+        [InlineData("event:name", true)]
+        public void Validate_ReadableId_AllowsOnlyAlphanumericHyphensUnderscores(string readableId, bool hasError)
```
ethax-ross, line 72: I think this will pass with a readable id like `event-name-` though not sure if its a big deal

ethax-ross (dismissed):

GetIntoTeachingApiTests/Models/GetIntoTeaching/Validators/TeachingEventUpsertOperationValidatorTests.cs
jkempster34: Pushed a change for this. I don't think the Regex is much worse.

ethax-ross (approved):

PR #814, opened 2021-10-19
Paginate Apply API sync job

ethax-ross:
[Trello-2551](https://trello.com/c/bGR36n8s/2551-update-apply-api-integration-to-support-pagination)

The Apply API now supports pagination by passing a `?page=<num>` query string attribute. The response headers include a `Total-Pages` and `Current-Page` key so that we can work out if there are further records to retrieve.

Create a new `PaginatorClient` that can take a `Flurl` resource and paginate according to the above specification. Update the `FindApplySyncJob` to use the new paginator class and queue all pages of candidates for syncing with the CRM.


peteryates (approved):
Looks good, makes sense.

PR #826, opened 2021-10-25
Temp cache candidate

jkempster34:
TODO:
- testing
- handle other candidate controller actions
- move some of the intermediate key stuff into new classes
- update schools experience - when they fetch a candidate, they can do it based on whether a CRM Guid is present - if it isn't present, they can use the new endpoint with the intermediate Guid

## 1. Add `Candidate` cache
   
However, in School Experience, the Candidate upsert cannot be queued because the candidate information is needed immediately by the Schools Experience app.

Add a `Candidate` cache and a intermediate table. The cache can be used when the CRM integration is paused. We can persist a temporary ID in the intermediate table which can be used to retrieve the CRM Candidate record once the cache has been cleared.

## 3. Queue candidate CRM job

We cache the `Candidate` while the CRM integration is paused. However, we don't want to persist this permanently as it contains personal information. 

Ensure that when the `Candidate` is inserted into the CRM, the API cache record is deleted. Due to the way the `CandidateUpserter` job is configured, the API record will always be deleted within 24 hours.

## 4. Add endpoint to get candidate by intermediate ID

If a candidate is created when the CRM is down, we can store the candidate intermediate ID in the School Experience app.

When we want to retrieve the record, we can check whether it still exists in the cache. If it does, return it, else lookup the CRM ID from the mapping table (this will populate the CRM ID for any further requests)

### DIFF
--- a/GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
+++ b/GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
@@ -25,17 +25,20 @@ namespace GetIntoTeachingApi.Controllers.SchoolsExperience
         private readonly ICrmService _crm;
         private readonly ICandidateUpserter _upserter;
         private readonly IBackgroundJobClient _jobClient;
+        private readonly IStore _store;
 
         public CandidatesController(
             ICandidateAccessTokenService tokenService,
             ICrmService crm,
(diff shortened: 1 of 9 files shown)


jkempster34:
Closing this because the approach has changed. Easier to redo and paste over the relevant bits.

PR #850, opened 2021-11-09
Cache candidate while CRM down

jkempster34:
## 1. Add delete method

We are going to cache the Candidate model temporarily while the CRM integration is paused. We will need to delete the cache when it is back up.

## 2. Add `Candidate` cache

However, in School Experience, the Candidate upsert cannot be queued because the candidate information is needed immediately by the Schools Experience app.

Add a Candidate cache. The cache can be used when the CRM integration is paused.

Note: This creates a table for each entity related to Candidate, despite not declaring a DbSet for them.

## 3. Cache `Candidate` in Store if CRM integration is paused

When the CRM integration is paused, we want to provide a GUID to the Candidate, and cache it.

Usually, it is best practice to allow the CRM to generate sequential GUIDs which provide better SQL performance. However, in this scenario we have agreed it is beneficial to provide the GUID up-front because the School Experience app needs the Candidate ID immediately.

The Candidate upsert job is also triggered, so that when the CRM is operational again, we can persist the stored Candidate there, deleting the cache.

## 4. Update `UpsertCandidateJob` to remove candidate cache when successful

We cache the `Candidate` while the CRM integration is paused. However, we don't want to persist this permanently as it contains personal information. 

Ensure that when the `Candidate` is inserted into the CRM, the API cache record is deleted. Due to the way the `CandidateUpserter` job is configured, the API record will always be deleted within 24 hours (which it needs to be because it contains personal data).

## 5. Update `Candidate` controller actions to use cache when CRM down

When the CRM integration is paused, the record may exist in the cache (if they have completed their sign up recently). When that is the case, fetch the record from there instead of the CRM.

# Thoughts
- Fetching from the cache won't work when the candidate has signed up **before** the CRM integration was paused. This is used for various stuff like cancelling a placement and accepting a placement. If we can't figure out a way to make the GETs work, we might as well get rid of the Store code.
- Linked to the above. I am querying the database/CRM conditionally based on whether the CRM integration is paused. There will be a period of time when it is unpaused, but that CandidateUpsertJob has not yet completed.

### DIFF
--- a/GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
+++ b/GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
@@ -25,17 +25,23 @@ namespace GetIntoTeachingApi.Controllers.SchoolsExperience
         private readonly ICrmService _crm;
         private readonly ICandidateUpserter _upserter;
         private readonly IBackgroundJobClient _jobClient;
+        private readonly IStore _store;
+        private readonly IDateTimeProvider _dateTime;
 
         public CandidatesController(
             ICandidateAccessTokenService tokenService,
             ICrmService crm,
             ICandidateUpserter upserter,
-            IBackgroundJobClient jobClient)
+            IBackgroundJobClient jobClient,
+            IStore store,
+            IDateTimeProvider dateTime)
         {
             _crm = crm;
             _upserter = upserter;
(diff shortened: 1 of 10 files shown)


jkempster34:
Went a bit off target here. No need to create a Candidate table. See here https://github.com/DFE-Digital/get-into-teaching-api/pull/851

PR #851, opened 2021-11-10
Cache candidate while CRM is down

jkempster34:
## Queue `CandidateUpsertJob` for sign up when CRM integration is paused

However, in School Experience, candidate information is needed immediately by the Schools Experience app. So, we must provide a GUID up-front to the Candidate before queuing the request. This way, the School Experience app can persist the CRM GUID when it receives a response (as it usually does when the CRM is operational).

Usually, it is best practice to allow the CRM to generate sequential GUIDs which provide better SQL performance. However, in this scenario we have agreed it is beneficial to provide the GUID up-front because the School Experience app needs the Candidate ID immediately.

## Put queuing candidate sign ups behind feature flag

We want to test this before enabling it in production.

### DIFF
--- a/GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
+++ b/GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
@@ -26,19 +26,22 @@ namespace GetIntoTeachingApi.Controllers.SchoolsExperience
         private readonly ICandidateUpserter _upserter;
         private readonly IBackgroundJobClient _jobClient;
         private readonly IDateTimeProvider _dateTime;
+        private readonly IEnv _env;
 
         public CandidatesController(
             ICandidateAccessTokenService tokenService,
             ICrmService crm,
             ICandidateUpserter upserter,
             IBackgroundJobClient jobClient,
-            IDateTimeProvider dateTime)
+            IDateTimeProvider dateTime,
+            IEnv env)
         {
             _crm = crm;
             _upserter = upserter;
             _tokenService = tokenService;
             _jobClient = jobClient;
             _dateTime = dateTime;
+            _env = env;
         }
 
         [HttpPost]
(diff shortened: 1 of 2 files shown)


GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
```diff
@@ -57,13 +63,30 @@ public class CandidatesController : ControllerBase
...
+            }
+            else
+            {
+                _upserter.Upsert(candidate);
             }
-
-            var candidate = request.Candidate;
-            _upserter.Upsert(candidate);
 
             return CreatedAtAction(
```
jkempster34, line 91: This should be an `AcceptedAtAction` status 202 https://stackoverflow.com/questions/37130993/what-should-be-the-http-status-code-for-request-pending

Not sure if it matter much

GetIntoTeachingApiTests/Controllers/SchoolsExperience/CandidatesControllerTests.cs
```diff
@@ -119,15 +125,34 @@ public void SignUp_ValidRequest_UpsertsCandidateAndRespondsWithSuccess()
...
         {
+            var request = new SchoolsExperienceSignUp { FirstName = "first" };
             var mockAppSettings = new Mock<IAppSettings>();
             mockAppSettings.Setup(m => m.IsCrmIntegrationPaused).Returns(true);
+
+            var response = _controller.SignUp(request, mockAppSettings.Object);
+
+            var created = response.Should().BeOfType<CreatedAtActionResult>().Subject;
+            var signUp = created.Value.Should().BeAssignableTo<SchoolsExperienceSignUp>().Subject;
+            signUp.FirstName.Should().Be(request.FirstName);
```
ethax-ross, line 138: Worth asserting that the `signUp.Id` is not null here? may also be worth adding a quick test around the production behaviour and/or a comment to say that its temporary while we roll it out

ethax-ross (approved):

PR #852, opened 2021-11-11
Add an `AddClassroomExperienceNoteJob` so that request can be queued

jkempster34:
Part of the School Experience candidate sign up (in the app) attempts to add a `ClassroomExperienceNote`. This firstly fetches the candidate from the CRM. 

When the CRM integration is paused, this is not possible and we were therefore seeing 404 not found.

To fix this, add an `AddClassroomExperienceNoteJob` so that the API can queue the request until the CRM is operational again. 

The `AddClassroomExperienceNoteJob` must complete after the `UpsertCandidateJob` of the `SignUp` action. However, the Hangfire jobs run concurrently, so we silently retry the `AddClassroomExperienceNoteJob` three times to allow some time for the `UpsertCandidateJob` to complete.

### DIFF
--- a/GetIntoTeachingApi/Jobs/AddClassroomExperienceNoteJob.cs
+++ b/GetIntoTeachingApi/Jobs/AddClassroomExperienceNoteJob.cs
@@ -1,0 +1,80 @@
+﻿using System;
+using GetIntoTeachingApi.Adapters;
+using GetIntoTeachingApi.Models;
+using GetIntoTeachingApi.Models.Crm;
+using GetIntoTeachingApi.Models.SchoolsExperience;
+using GetIntoTeachingApi.Services;
+using GetIntoTeachingApi.Utils;
+using Hangfire;
+using Hangfire.Server;
+using Microsoft.Extensions.Logging;
+
+namespace GetIntoTeachingApi.Jobs
+{
+    public class AddClassroomExperienceNoteJob : BaseJob
+    {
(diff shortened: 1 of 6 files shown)


GetIntoTeachingApi/Jobs/AddClassroomExperienceNoteJob.cs
```diff
@@ -0,0 +1,79 @@
...
+                throw new InvalidOperationException($"{GetType().Name} - Aborting (CRM integration paused).");
+            }
+
+            var existingCandidate = _crm.GetCandidate(candidateId);
+
+            if (existingCandidate == null)
+            {
+                if (CurrentAttempt(context, _contextAdapter) <= NumberOfQuietRetries)
+                {
+                    _logger.LogInformation($"{GetType().Name} - Candidate not found (may be in concurrent job queue)");
```
ethax-ross, line 55: I don't think this is going to retry the job if it hits this conditional; the job will complete?

ethax-ross (commented):
Looks good, I don't think the job will retry properly though it looks like it will just finish successfully in that scenario

ethax-ross (approved):
Looks good to me 👍  shame about having to do all that custom mocking 😩 I wonder if it would be possible/simpler to instead check the queue state through `JobStorage.Current.GetMonitoringApi()` to check if its been requeued... I don't mind either way though, looks good.

PR #855, opened 2021-11-12
Fix candidate Id not being set before being sent to job

jkempster34:
I hadn't updated the candidate sent to the `UpsertCandidateJob`. The candidate was being sent to the CRM with a `null` ID and the CRM was generating one.

Also fix SonarCloud code smell that was left in from refactoring.


ethax-ross (approved):
Well spotted 👍

PR #860, opened 2021-11-15
Ensure correct method is called when the ID is API generated

jkempster34:
Currently, the `CrmService` chooses the which `OrganisationService` method to call based on whether the model has an existing ID.

However, in certain circumstances, we need to allow to API to generate an ID up-front. 

Add a `GeneratedUpfrontId` method to the `BaseModel`, which sets a flag of `HasUpfrontId` on the model to true. The `BaseModel` can then choose which `OrganisationService` method to call based on the value of the flag.

### DIFF
--- a/GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
+++ b/GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
@@ -4,7 +4,6 @@ using System.Linq;
 using GetIntoTeachingApi.Attributes;
 using GetIntoTeachingApi.Jobs;
 using GetIntoTeachingApi.Models;
-using GetIntoTeachingApi.Models.Crm;
 using GetIntoTeachingApi.Models.SchoolsExperience;
 using GetIntoTeachingApi.Services;
 using GetIntoTeachingApi.Utils;
@@ -76,7 +75,7 @@ namespace GetIntoTeachingApi.Controllers.SchoolsExperience
                 // Usually, it is best practice to allow the CRM to generate sequential GUIDs which provide better
                 // SQL performance. However, in this scenario we have agreed it is beneficial to provide the GUID up-front
                 // because the School Experience app needs the Candidate ID immediately.
-                candidate.Id = Guid.NewGuid();
+                candidate.GenerateUpfrontId();
 
                 // This is the only way we can mock/freeze the current date/time
                 // in contract tests (there's no other way to inject it into this class).
--- a/GetIntoTeachingApi/Models/Crm/BaseModel.cs
+++ b/GetIntoTeachingApi/Models/Crm/BaseModel.cs
@@ -7,32 +7,35 @@ using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.Serialization;
-using System.Text.Json.Serialization;
 using FluentValidation;
 using GetIntoTeachingApi.Attributes;
 using GetIntoTeachingApi.Services;
 using Microsoft.Xrm.Sdk;
 using Microsoft.Xrm.Sdk.Client;
+using Newtonsoft.Json;
 
 namespace GetIntoTeachingApi.Models.Crm
 {
-    public class BaseModel : INotifyPropertyChanged
+    public abstract class BaseModel : INotifyPropertyChanged
     {
-        private readonly string[] _propertyNamesExcludedFromChangeTracking = new string[] { "ChangedPropertyNames" };
+        private readonly string[] _propertyNamesExcludedFromChangeTracking = new string[] { nameof(ChangedPropertyNames), nameof(HasUpfrontId) };
         private bool _changeTrackingEnabled = true;
 
         [NotMapped]
-        [JsonIgnore]
+        [JsonProperty]
+        public bool HasUpfrontId { get; private set; }
+        [NotMapped]
+        [System.Text.Json.Serialization.JsonIgnore]
         public HashSet<string> ChangedPropertyNames { get; set; } = new HashSet<string>();
         [DatabaseGenerated(DatabaseGeneratedOption.None)]
         public Guid? Id { get; set; }
 
-        public BaseModel()
+        protected BaseModel()
         {
             InitChangedPropertyNames();
         }
 
-        public BaseModel(Entity entity, ICrmService crm, IValidatorFactory vaidatorFactory)
+        protected BaseModel(Entity entity, ICrmService crm, IValidatorFactory vaidatorFactory)
             : this()
         {
             Id = entity.Id;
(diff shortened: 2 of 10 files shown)


GetIntoTeachingApi/Services/ICrmService.cs
```diff
@@ -32,6 +32,7 @@ public interface ICrmService
         void DeleteLink(Entity source, Relationship relationship, Entity target, OrganizationServiceContext context);
         IEnumerable<Entity> RelatedEntities(Entity entity, string relationshipName, string logicalName);
         Entity MappableEntity(string entityName, Guid? id, OrganizationServiceContext context);
+        Entity NewEntity(string entityName, OrganizationServiceContext context);
```
ethax-ross, line 35: Can we make the interface more explicit and have `NewEntity` and `BlankExistingEntity` as methods here, instead of retaining the `MappableEntity` which in some cases will also call `NewEntity`?

GetIntoTeachingApi/Models/Crm/BaseModel.cs
```diff
@@ -148,6 +150,11 @@ private static string TrimAndNullifyIfEmpty(string input)
             return string.IsNullOrWhiteSpace(input) ? null : input;
         }
 
+        private bool HasGeneratedUpfrontId()
+        {
+            return this is ICanGenerateUpfrontId && ((ICanGenerateUpfrontId)this).ApiOnlyId;
```
ethax-ross, line 155: I think we should have `BaseModel` implement `ICanGenerateUpfrontId` instead of `Candidate` - it would make sense to have `HasGeneratedUpfrontId` on the interface as well

GetIntoTeachingApi/Models/Crm/ICanGenerateUpfrontId.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace GetIntoTeachingApi.Models.Crm
+{
+    public interface ICanGenerateUpfrontId
+    {
+        public Guid? Id { get; set; }
+        public bool ApiOnlyId { get; set; }
+
+        public void GenerateId()
```
ethax-ross, line 10: TIL default implementations are a thing 🤯

GetIntoTeachingApi/Models/Crm/ICanGenerateUpfrontId.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace GetIntoTeachingApi.Models.Crm
+{
+    public interface ICanGenerateUpfrontId
+    {
+        public Guid? Id { get; set; }
+        public bool ApiOnlyId { get; set; }
```
ethax-ross, line 8: This property reads funny to me; would it make more sense/link to the interface better as `HasUpfrontId`?

GetIntoTeachingApiTests/Controllers/SchoolsExperience/CandidatesControllerTests.cs
```diff
@@ -240,12 +240,14 @@ public void AddClassroomExperienceNote_ValidRequest_EnqueuesJobAndRespondsWithNo
                 It.IsAny<EnqueuedState>()));
         }
 
-        private static bool IsRequestPlusId(Candidate requestCandidate, string candidateSentToJobJson, Guid expectedId)
+        private static bool IsRequestPlusIdAndApiGeneratedIdFlag(Candidate requestCandidate, string candidateSentToJobJson, Guid expectedId)
```
ethax-ross, line 243: I re-read this method name a few times 😅 how about something like `MatchesCandidateWithUpfrontId`?

GetIntoTeachingApiTests/Models/Crm/BaseModelTests.cs
```diff
@@ -332,6 +332,17 @@ public void ToEntity_WithNew()
             mockEntity.GetAttributeValue<string>("dfe_field3").Should().Be(mock.Field3);
         }
 
+        [Fact]
+        public void ToEntity_WhenIdIsApiGenerated_CallsNewEntity()
+        {
+            var mock = new ICanGenerateUpfrontIdMock();
```
ethax-ross, line 338: If we have `BaseModel` implement the interface I think we could just use that model directly, or one of the subclasses to avoid having to mock it?

ethax-ross (commented):
I can't think of a better general approach to this; I think it could be tweaked here or there though to make it a bit simpler/easier to follow

GetIntoTeachingApiTests/Models/Crm/ICanGenerateUpfrontIdTests.cs
```diff
@@ -0,0 +1,35 @@
...
+            var mock = new MockModel { Id = Guid.NewGuid() };
+
+            Action action = () => ((ICanGenerateUpfrontId)mock).GenerateId();
+
+            action.Should().Throw<InvalidOperationException>()
+                .WithMessage("Can only generate IDs on models that do not have a preexisting ID.");
+        }
+
+        [Fact]
+        public void GenerateId_WhenNoPreexistingId_GeneratesIdAndSetsApiOnlyIdFlagToTrue()
```
ethax-ross, line 23: `GeneratesIdAndSetsHasUpfrontIdFlagToTrue`

GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
```diff
@@ -76,7 +76,7 @@ public class CandidatesController : ControllerBase
                 // Usually, it is best practice to allow the CRM to generate sequential GUIDs which provide better
                 // SQL performance. However, in this scenario we have agreed it is beneficial to provide the GUID up-front
                 // because the School Experience app needs the Candidate ID immediately.
-                candidate.Id = Guid.NewGuid();
+                ((ICanGenerateUpfrontId)candidate).GenerateId();
```
ethax-ross, line 79: Do we need the casts to the interface here and on `MockModel` any more since `BaseModel` implements it?

GetIntoTeachingApiTests/Models/Crm/BaseModelTests.cs
```diff
@@ -332,6 +332,17 @@ public void ToEntity_WithNew()
             mockEntity.GetAttributeValue<string>("dfe_field3").Should().Be(mock.Field3);
         }
 
+        [Fact]
+        public void ToEntity_WhenIdIsApiGenerated_CallsNewEntity()
```
ethax-ross, line 336: `WhenIdIsGeneratedUpfront` for consistency?

ethax-ross (commented):
Looks good, just a couple of minor suggestions

ethax-ross (approved):
Looks good 👍

PR #875, opened 2021-11-30
Relax event readable id validation

ethax-ross:
[Trello-2686](https://trello.com/c/4JnItmB8/2686-investigate-400-error-on-internal-events-form)

The `ReadableId` is used as the URL slug for an event; a number of pre-existing events have invalid ids according to our existing regex used for validation. This causes issues when these events are withdrawn/edited in the app. To avoid replacing all of these slugs with valid ones and setting up redirects we're instead going to relax the validation in the API - whilst  retaining stricter validation in the app - until the events are in the past. When they are, we should be good to reinstate the stricter regex (the last event with an invalid slug is currently 24/6/21).


GetIntoTeachingApi/Models/GetIntoTeaching/Validators/TeachingEventUpsertOperationValidator.cs
```diff
@@ -12,10 +12,14 @@ public TeachingEventUpsertOperationValidator(ICrmService crm)
         {
             _crm = crm;
 
+            // The regex we use here has been relaxed be due to a number of
+            // pre-existing events that would otherwise fail the validation.
+            // We should look to revert it back to the stricter regex in the
+            // future (when these events have since passed - 24/6/21).
```
jkempster34, line 18: I think this should be 2022. Wondering whether it's worth putting a `TODO: ` in so that this is picked up by SonarCloud and not left to be forgotten?

ethax-ross: Whoops. Yep.

jkempster34 (approved):
LGTM. Thanks for fixing 🙇‍♂️

PR #876, opened 2021-12-01
Label TOTP metrics with service reference

ethax-ross:
[Trello-2692](https://trello.com/c/HYJoDiVp/2692-investigate-if-authentication-is-causing-significant-drop-off-on-event-sign-ups)

We want to be able to attribute TOTP-related metrics to individual services (and perhaps later to individual transactions within a service).

- Add Reference attribute to ExistingCandidateRequest

We want to be able to attribute authentication actions (generate TOTP/verify TOTP) to specific transactions; in order to do this we will get the clients to pass a reference (e.g. "Mailing List", "Event", etc).

- Add reference label to TOTP metrics

We currently track the generated/verified TOTPs; we now want to break this down by service transaction.

Add a label that can store a reference for the TOTP action and populate it in the token generation service.

- Add a new claim to store the client name

We can currently only retrieve the token/role of the authenticated client, however we need to be able to retrieve the name of the client so that we can later use it when attributing TOTP actions that occur. This will enable us to break down the TOTP actions by service (we can later make this more granular by passing the transaction name from the clients).

- Populate ExistingCandidateRequest reference

As the clients do not yet send a reference explicitly we are going to default them to the client name (which will give us a reasonable breakdown without having to update each client). If we find this is not granular enough we can update the clients to provide a reference relative to the individual transactions.

### DIFF
--- a/GetIntoTeachingApi/Controllers/GetIntoTeaching/MailingListController.cs
+++ b/GetIntoTeachingApi/Controllers/GetIntoTeaching/MailingListController.cs
@@ -82,6 +82,8 @@ namespace GetIntoTeachingApi.Controllers.GetIntoTeaching
             [FromRoute, SwaggerParameter("Access token (PIN code).", Required = true)] string accessToken,
             [FromBody, SwaggerRequestBody("Candidate access token request (must match an existing candidate).", Required = true)] ExistingCandidateRequest request)
         {
+            request.Reference ??= User.Identity.Name;
+
             var candidate = _crm.MatchCandidate(request);
 
             if (candidate == null || !_accessTokenService.IsValid(accessToken, request, (Guid)candidate.Id))
--- a/GetIntoTeachingApiTests/Helpers/ControllerBaseExtensions.cs
+++ b/GetIntoTeachingApiTests/Helpers/ControllerBaseExtensions.cs
@@ -1,0 +1,20 @@
+﻿using System.Security.Claims;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+
+namespace GetIntoTeachingApiTests.Helpers
+{
+    public static class ControllerBaseExtensions
+    {
+        public static void MockUser(this ControllerBase controller, string name = "Client")
+        {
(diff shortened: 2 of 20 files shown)


GetIntoTeachingApiTests/Helpers/ControllerBaseExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Security.Claims;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+
+namespace GetIntoTeachingApiTests.Helpers
+{
+    public static class ControllerBaseExtensions
+    {
+        public static void MockUser(this ControllerBase controller, string Name = "Client")
```
jkempster34, line 9: Think this `Name` parameter should be lowercase

GetIntoTeachingApi/Controllers/GetIntoTeaching/MailingListController.cs
```diff
@@ -82,6 +82,8 @@ public class MailingListController : ControllerBase
             [FromRoute, SwaggerParameter("Access token (PIN code).", Required = true)] string accessToken,
             [FromBody, SwaggerRequestBody("Candidate access token request (must match an existing candidate).", Required = true)] ExistingCandidateRequest request)
         {
+            request.Reference ??= User.Identity.Name;
```
jkempster34, line 85: Could we not set the transaction name here `request.Reference = "Mailing list"`? Same question for the other transaction specific actions.

jkempster34 (changes requested):
Looks good! Found some minor stylistic stuff, and have one question in there too

GetIntoTeachingApi/Controllers/GetIntoTeaching/MailingListController.cs
ethax-ross: Yeah that's going to be phase two if we find that we need that granularity as it will need a bit of work in the clients to ensure we're sending the transaction-level reference on the create access token request; its going to be easier to dashboard it in Grafana if its all at the service-level for now

jkempster34 (approved):

PR #878, opened 2021-12-03
Take latest application form for contact-level

ethax-ross:
[Trello-2729](https://trello.com/c/XaOYBfWz/2729-apply-candidate-sync-job-bug-fix)

The Apply API should be returning application forms in order of most recently updated first, however this doesn't appear to be the case. Instead, we're going to order the application forms explicitly in the API, taking the latest phase/status for the contact-level attributes.


jkempster34 (approved):
👍 Are these tests relying on the default value of Datetime being 1970? Is it worth explicitly assigning a Datetime to all of the forms

ethax-ross:
> 👍 Are these tests relying on the default value of Datetime being 1970? Is it worth explicitly assigning a Datetime to all of the forms

Yeah good shout; I think the `UpdatedAt` field used to be nullable but its not any more - I'll fixup the change.

jkempster34 (approved):

PR #879, opened 2021-12-07
Se entity

jkempster34:
- Add new CandidateSchoolExperience model

Currently, we add a ClassroomExperienceNote when the status of a School Experience candidate changes (e.g., they make a placement request, their placement is accepted etc.). This is a single formatted text field due to some time constraints during initial development.

The CRM team have created a new table - CandidateSchoolExperience - to represent this data. 

Add the initial models and validators. For now, the ClassroomExperienceNote will stay in place until the new feature is verified. 

- Add new endpoint for CandidatesSchoolExperience

To enable us to switch over to using the new CandidatesSchoolExperience model, we are going to make a new endpoint available in non-production environments. When the feature is verified we will remove the AddClassroomExperienceNotes endpoint.

This all seems to be working properly. Here is a candidate with some school experience records that I have added through the new endpoint
image


GetIntoTeachingApi/Models/Crm/Candidate.cs
```diff
@@ -301,9 +301,10 @@ public enum RegistrationStatus
         [SwaggerSchema("Set to update the accepted privacy policy.", WriteOnly = true)]
         [EntityRelationship("dfe_contact_dfe_candidateprivacypolicy_Candidate", typeof(CandidatePrivacyPolicy))]
         public CandidatePrivacyPolicy PrivacyPolicy { get; set; }
+        [EntityRelationship("dfe_candidateschoolexperience", typeof(CandidateSchoolExperience))]
+        public List<CandidateSchoolExperience> SchoolExperiences { get; set; } = new List<CandidateSchoolExperience>();
 
         public Candidate()
-            : base()
```
ethax-ross, line 306: What's the reasoning behind this change? the `BaseModel` constructor has a purpose looking at this, so just want to make sure its intentional?

https://github.com/DFE-Digital/get-into-teaching-api/blob/master/GetIntoTeachingApi/Models/Crm/BaseModel.cs#L36

GetIntoTeachingApi/Models/Crm/CandidateSchoolExperience.cs
```diff
@@ -0,0 +1,50 @@
...
+        [EntityField("statuscode")]
+        public int Status { get; set; }
+        [EntityField("dfe_teachingsubject", typeof(EntityReference), "dfe_teachingsubjectlist")]
+        public Guid? TeachingSubjectId { get; set; }
+        [EntityField("dfe_notes")]
+        public string Notes { get; set; }
+        [EntityField("dfe_where")]
+        public string SchoolName { get; set; }
+
+        public CandidateSchoolExperience()
```
ethax-ross, line 41: Again, just checking the lack of `base()` call here is intentional?

GetIntoTeachingApiTests/Models/Crm/Validators/CandidateSchoolExperienceValidatorTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                TeachingSubjectId = Guid.NewGuid(),
+            };
+
+            var result = _validator.TestValidate(candidateSchoolExperience);
+
+            result.ShouldHaveValidationErrorFor(c => c.TeachingSubjectId);
+        }
+
+        [Fact]
+        public void Validate_Notes_HasError()
```
ethax-ross, line 92: Worth being consistent with the method names on these to elsewhere? so something like `Validate_WhenNotesIsTooLong_HasError` etc

GetIntoTeachingApi/Models/Crm/Validators/CandidateSchoolExperienceValidator.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public class CandidateSchoolExperienceValidator : AbstractValidator<CandidateSchoolExperience>
+    {
+        private readonly IStore _store;
+
+        public CandidateSchoolExperienceValidator(IStore store)
+        {
+            _store = store;
+
+            RuleFor(se => se.SchoolUrn).MaximumLength(8);
```
ethax-ross, line 15: Will we get a validation error on `CandidateId` when its not specified given its non-nullable, or does it default to an empty guid of 0's? I see to remember having to make these `Guid?` and have a `NotNull()` rule for them to behave, but I may be misremembering

GetIntoTeachingApi/Models/Crm/Candidate.cs
jkempster34: Just following Visual Studio, did some reading and seems like you don't need that call to Base unless you are passing arguments. I'll put them back in to be consistent though
image

GetIntoTeachingApi/Models/Crm/Validators/CandidateSchoolExperienceValidator.cs
jkempster34: You may be misremembering, I have copied off the ApplicationForm model and that doesn't do this

ethax-ross (dismissed):
Looks great 👍

ethax-ross (approved):

ethax-ross (dismissed):

ethax-ross (approved):

PR #881, opened 2021-12-10
Fix various linting warnings

ethax-ross:
[Trello-2742](https://trello.com/c/WMMOuQeh/2742-improve-linting-in-the-api)

SonarCloud has highlighted a number of code issues, namely:

- Methods that could be static
- Private classes that should be `sealed`
- Use of Array.Empty
- Logging message template varying between calls
- LINQ tweaks/avoiding unnecessary `Count()`
- An enum with the same values
- Exceptions adhearing to `ISerializable`
- Calling `GC.SuppressFinalize` on `Dispose`

There are a few more to look at but I'll address those separately.


jkempster34 (approved):
👍

PR #883, opened 2021-12-10
Configure NETAnalyzer static code analysis

ethax-ross:
Configures the project with NETAnalyzer static code analysis (Microsoft's static code analyzer) and fixes all the recommended violations.

We can take this a step further with:

```
<AnalysisMode>Recommended</AnalysisMode>
```

But it flags up around 300 violations and there doesn't appear to be a solution-wide auto-fix. [I can't face that on a Friday](https://trello.com/c/tEmKXBtL/2754-enable-and-fix-all-linting-errors-in-the-api).


GetIntoTeachingApi/Jobs/BaseJob.cs
```diff
@@ -6,7 +6,7 @@ namespace GetIntoTeachingApi.Jobs
 {
     public abstract class BaseJob
     {
-        protected readonly IEnv Env;
+        internal readonly IEnv Env;
```
jkempster34, line 9: Looks good, but the protection level here is worse. Wonder if this could be an auto property along the lines of `protected  IEnv Env { get; init; }`

jkempster34: Just noticed that the casing is wrong, wonder if that is what is upsetting the static analyser. I think its recommended to use auto properties where possible anyway.

jkempster34 (approved):
LGTM! 💯

PR #897, opened 2021-12-22
Updated documentation and system diagrams

ethax-ross:
[Trello-2681](https://trello.com/c/RzwGp03i/2681-up-to-date-architecture-diagrams)

- Fix markdown formatting and restore links

The Markdown formatting was a bit off on some of these. Also restores some links that were in the Confluence docs these were migrated from.

- Add architecture documentation

Add an architecture document that includes a system diagram of the Get into Teaching systems. I'm not sure the best place for this to live, but we seem to bunch other things that are GiT-wide into the API repo so it probably makes sense for now.

Uses [diagrams.net](https://diagrams.net) - import `architecture.drawio` to modify.

---

architecture


jkempster34 (commented):
This looks cool and imported fine in diagrams.net. Just wondering if this part looks like the apps are sharing the same databases, rather than the diagram just reusing the symbol? 

image

ethax-ross:
@jkempster34 the apps do share the same Postgres instance; I think it was a faff/expensive to have multiple high-availability Postgres instances so the way its setup they all share the same one, albeit using different databases for Postgres and namespaces for Redis

jkempster34 (approved):
> @jkempster34 the apps do share the same Postgres instance; I think it was a faff/expensive to have multiple high-availability Postgres instances so the way its setup they all share the same one, albeit using different databases for Postgres and namespaces for Redis

Didn't know that! Don't know if it is worth adding GOV Notify and Google Geocoding to School Experience too, but looks nice 💯

ethax-ross:
Ah yeah that's a good shout, I'll get this merged down then do that as another PR later on

PR #899, opened 2022-01-05
Revert endpoint and add primary reserve numbers

jkempster34:
- Revert "Get candidates with different primary and secondary phone numbers"

- Use other phone numbers when primary is null

We have simplified the School Experience sign up by removing alternate phone numbers. However, there are a number of records in the CRM that have School Experience (they have a Classroom Experience Note), but no primary phone number. 

If any of these candidates have open requests, they will not have a number in the Manage section of School Experience. Any new sign ups write to primary number so won't be a problem.

To fix this, attempt to fetch (in order) mobile, address, and secondary phone numbers from the Candidate model if primary number is null. The CRM team are going to fill the blanks, so this may be able to be removed shortly.


GetIntoTeachingApi/Models/SchoolsExperience/SchoolsExperienceSignUp.cs
```diff
@@ -66,7 +66,12 @@ private void PopulateWithCandidate(Candidate candidate)
             AddressCity = candidate.AddressCity;
             AddressStateOrProvince = candidate.AddressStateOrProvince;
             AddressPostcode = candidate.AddressPostcode;
-            Telephone = candidate.Telephone.StripExitCode();
+
+            // We only want to expose primary phone number to School Experience. However, there
+            // are a number of records in the CRM that have School Experience but no primary number.
+            // Fetch the other phone numbers for now until the issue has been resolved in the CRM.
+            var telephoneReserves = new string[] { candidate.MobileTelephone, candidate.AddressTelephone, candidate.SecondaryTelephone };
+            Telephone = candidate.Telephone.StripExitCode() ?? Array.Find(telephoneReserves, number => !string.IsNullOrWhiteSpace(number));
```
ethax-ross, line 74: Would we not want to call `StripExitCode()` on the result of this expression, or is it intentional it only applies to `candidate.Telephone` and not the others?

ethax-ross (approved):
Makes sense

ethax-ross (approved):

PR #902, opened 2022-01-10
Ignore assets in HighNotFound app alert

ethax-ross:
We appear to get a lot of traffic from bots/crawlers that hit non-hashed versions of the assets and get a 404. As we check all of our internal assets in our unit tests I think we should be safe to exclude them here; 404s are likely only to come from external websites that have linked to an image we have since removed, which we can do nothing about.

Exclude `*.txt` files as we don't host any and its another source of bot traffic (i.e. `passwords.txt` 🤦).


peteryates (approved):
Makes sense to me

PR #919, opened 2022-02-10
New email template verification code

ethax-ross:
- Use new email template on create access token

The new email template has updated copy and uses the candidate's first name to make it more personalised.

Screenshot 2022-02-21 at 08 25 28

### DIFF
--- a/GetIntoTeachingApi/Controllers/CandidatesController.cs
+++ b/GetIntoTeachingApi/Controllers/CandidatesController.cs
@@ -82,7 +82,10 @@ namespace GetIntoTeachingApi.Controllers
             }
 
             var token = _accessTokenService.GenerateToken(request, (Guid)candidate.Id);
-            var personalisation = new Dictionary<string, dynamic> { { "pin_code", token } };
+            var personalisation = new Dictionary<string, dynamic> {
+                { "pin_code", token },
+                { "first_name", request.FirstName }
+            };
 
             // We respond immediately/assume this will be successful.
             _notifyService.SendEmailAsync(request.Email, NotifyService.NewPinCodeEmailTemplateId, personalisation);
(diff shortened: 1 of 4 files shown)


SaimaBodiat:
I think it's best if we strip back and make this simpler and more general since it has to apply to several services. I suggest we tweak the original copy Ellie wrote.

“Ross, you’ve received this email because you’re already registered with us.

To confirm your details…

This code is only valid for 15 minutes. If your code expires, you can request a new on the page you came from. 

Kind regards,

Get Into Teaching “

@EllieNodder let me know your thoughts -  'on the page you came from' might need a bit of content work

SaimaBodiat:
just to add, we can sign off as 'Get Into teaching' for all services. The Get School Experience page also refers to GIT for support so it should be fine.

EllieNodder:
> I think it's best if we strip back and make this simpler and more general since it has to apply to several services. I suggest we tweak the original copy Ellie wrote.
> 
> “Ross, you’ve received this email because you’re already registered with us.
> 
> To confirm your details…
> 
> This code is only valid for 15 minutes. If your code expires, you can request a new on the page you came from.
> 
> Kind regards,
> 
> Get Into Teaching “
> 
> @EllieNodder let me know your thoughts - 'on the page you came from' might need a bit of content work

**New suggested copy for all three services (Get Into Teaching/Get School Experience/Get an adviser)**

Saima, you’ve received this email because you’re already registered with us.

To confirm your details, please enter your verification code on the sign up page.

Your verification code:

**843413**

This code is only valid for 15 minutes. If it expires, you can request a new one on the sign up page where you enter your code.

Kind regards,

Get Into Teaching


@gemmadallmandfe @SaimaBodiat @ethax-ross

SaimaBodiat:
> 

Looks good to me!

MylesJarvis:
Looks great ✅ 
cc @gemmadallmandfe @ethax-ross @SaimaBodiat

jkempster34 (approved):

PR #920, opened 2022-02-14
Deduplicate UpsertCandidateJob

ethax-ross:
[Trello-2952](https://trello.com/c/MuyRT0qK/2952-investigate-duplicate-records-reported-in-crm)

- Deduplicate UpsertCandidateJobs 

We occasionally find in the CRM that duplicate records have been created (i.e. multiple records with the same first name, last name and email address).

There is already quite a lot of protection in to prevent this; the end-services disable form submission buttons to prevent double-clicks and the API splits updating a candidate into several jobs in case one of the steps/requests fails (in which case only that step/request is retried).

That being said, there are still ways for duplicates to be created; the only one I'm aware of at the moment is if Javascript is disabled then a user can submit a form multiple times by double-clicking the submit button. There's nothing we can do about this, as the only prevention requires Javascript.

In an effort to deduplicate records in a more centralised way we are going to store a signature in Redis for `UpsertCandidateJob`s are they are processed; that way, if a subsequent job begins processing and finds the signature already in Redis it can de-duplicate itself by completing early.

The redis keys will have a short TTL of 5 seconds to ensure legitimate jobs that are either the same or very similar should still run.

Initially I wanted to use a hash of the entire JSON job payload as the signature, however due to the way the requests are processed these can contain dates set to 'now', which will change slightly between duplicate jobs. Instead, I've opted to make the signature:

``` 
base_job.<job-type>.<candidate-id>-<candidate-email>-<changed-properties>
```

This should lead to a unique enough key, assuming two distinct transactions/endpoints don't accept the same set of attributes which is certainly the case at the moment.

- Update other jobs to adhere to BaseJob interface

The `IRedisService` parameter is now required by all constructors.

### DIFF
--- a/GetIntoTeachingApi/Jobs/BaseJob.cs
+++ b/GetIntoTeachingApi/Jobs/BaseJob.cs
@@ -1,4 +1,8 @@
-﻿using GetIntoTeachingApi.Adapters;
+﻿using System;
+using System.Security.Cryptography;
+using System.Text;
+using GetIntoTeachingApi.Adapters;
+using GetIntoTeachingApi.Services;
 using GetIntoTeachingApi.Utils;
 using Hangfire.Server;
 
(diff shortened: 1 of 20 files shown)


peteryates (approved):
Looks sensible. Still feels like this should be happening inside the CRM though!

PR #928, opened 2022-02-24
Improve Network Policies

sfawcett123:
Use terraform to manage network policies


terraform/paas/variables.tf
```diff
@@ -37,6 +37,10 @@ variable "paas_space" {
   default = "sandbox"
 }
 
+variable "paas_monitoring_app" {
+  default = "prometheus-dev-get-into-teaching"
```
saliceti, line 41: If we always expect to pass a value, we can remove the default

sfawcett123: changed

saliceti (approved):

PR #941, opened 2022-03-09
Remove Cypress tests on deploy

ethax-ross:
These have been migrated into the individual apps; at the moment there's no way of running them independently, but if we feel they are useful we can expose them from the app/tta and call them from the API build pipeline.

I think the contract tests are probably enough coverage, but we may want to create contracts for the mailing list/event sign ups in addition to the TTA.


peteryates (approved):
bye

PR #942, opened 2022-03-09
Revert "Update various dependencies"

ethax-ross:
Reverts DFE-Digital/get-into-teaching-api#936

In testing the sign ups functioned correctly but the hangfire dashboard isn't behaving; I think this is due to the change in postgres provider; going to revert while I investigate.


peteryates (approved):
Looks good

PR #944, opened 2022-03-09
Update various dependencies

ethax-ross:
Second attempt at #936 

- Migrate Hangfire database to new storage provider

Our new Postgres storage provider (compatible with Npgsql v6) is not compatible with the structure from the old provider. Its difficult to work out what the difference is in order to create a migration to fix it (as they are very similar) so we can instead point the new provider to a separate schema, which also means we should be able to revert without issues.

We'll need to manually check the job queue after releasing to clear any jobs that are pending down (from the old provider schema).

### DIFF
--- a/GetIntoTeachingApi/Database/DbConfiguration.cs
+++ b/GetIntoTeachingApi/Database/DbConfiguration.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.Linq;
 using System.Text.Json;
 using System.Text.Json.Serialization;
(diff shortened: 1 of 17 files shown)


GetIntoTeachingApi/Database/DbConfiguration.cs
```diff
@@ -35,6 +37,54 @@ public void Migrate()
...
+                "DROP TABLE hangfire.state;" +
+                "DROP TABLE hangfire.jobparameter;" +
+                "DROP TABLE hangfire.jobqueue;" +
+                "DROP TABLE hangfire.server;" +
+                "DROP TABLE hangfire.list;" +
+                "DROP TABLE hangfire.set;" +
+                "DROP TABLE hangfire.lock;" +
+                "DROP TABLE hangfire.counter;" +
+                "DROP TABLE hangfire.hash;" +
+                "DROP SCHEMA hangfire Cascade;";
```
peteryates, line 79: does this leave anything remaining in the `hangfire` schema? If not wouldn't dropping the schema take all the contents with it?

peteryates (approved):
Looks fine overall, bit dramatic! 😅

peteryates (approved):
Much nicer!

PR #945, opened 2022-03-14
Fix LocationSyncJob for when CSV header row is missing

ethax-ross:
[Trello-3036](https://trello.com/c/zYhSX7t7/3036-fix-api-location-sync-job-the-csv-format-has-changed)

The data source we use for a CSV of UK postcodes has updated their format so that it omits the header row and contains malformed rows. I'm not sure if the header being omitted has been an intentional change, so we need to account for if the row is present or omitted and if the row is valid or not.

Check the rows for validity and skip if malformed (the header row is treated as malformed and also skipped if present).


jkempster34 (approved):
✅

jkempster34 (approved):

PR #958, opened 2022-04-07
Add alerts for a high number of CSP violations

ethax-ross:
Add alerts to flag if we get an unusually high number of CSP violations (>20 in a 5 minute period). This should weed out CSP violations caused by ad-blockers for individual clients and (hopefully) only alert if multiple users are receiving the same CSP violation.

Suffix alert names with the app to make it easier to distinguish where the alert originated from (this is mainly useful for the app/tta alerts as they have the same names in most cases).

### DIFF
--- a/monitoring/prometheus/alert.rules
+++ b/monitoring/prometheus/alert.rules
@@ -39,7 +39,7 @@ groups:
           dashboard: https://grafana-prod-get-into-teaching.london.cloudapps.digital/d/eF19g4RZx/cf-apps?orgId=1
   - name: App
     rules:
-      - alert: TooManyRequests
+      - alert: TooManyRequests (App)
         expr: 'sum(increase(app_requests_total{path!~"csp_reports",status=~"429"}[1m])) > 0'
         labels:
           severity: high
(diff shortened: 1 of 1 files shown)


monitoring/prometheus/alert.rules
```diff
@@ -75,9 +75,18 @@ groups:
...
           runbook: https://dfedigital.atlassian.net/wiki/spaces/GGIT/pages/2156036281/App+TTA+Runbook#HighMemory-MEDIUM
           dashboard: https://grafana-prod-get-into-teaching.london.cloudapps.digital/d/0PRnzc2Mk/get-into-teaching-apps?viewPanel=33&orgId=1
+      - alert: HighCspViolations (App)
+        expr: 'sum(increase(app_csp_violations_total{app="get-into-teaching-app-prod"}[5m])) by (blocked_uri) > 20'
+        labels:
+          severity: medium
+        annotations:
+          summary: Alerts when more than 20 CSP violations for the same URI are reported in a 5 minute period.
+          description: Alerts when more than 20 CSP violations for the same URI are reported in a 5 minute period.
+          runbook: https://dfedigital.atlassian.net/wiki/spaces/GGIT/pages/2156036281/App+TTA+Runbook#HighCspViolations-MEDIUM
```
peteryates, line 85: didn't even know we had this 🙈

peteryates (approved):

PR #962, opened 2022-04-22
Updated CRM models for Apply API v1.2

ethax-ross:

- Add CRM models for v1.2 of Apply API integration

The Apply API integration v1.2 has additional models related to the `ApplicationForm` as well as a few new fields on the form itself.

Add `IHasFindApplyId` interface which will be useful later in creating generic methods for Find/Apply models.

Update `OperationsControllerTests` to reflect additional CRM model mappings.

As these models/fields are only available in the CRM dev environment we are adding a new feature flag to ensure they don't get accessed in environments that don't yet support them (we only have to do this for the attributes on the form model as the relationships will always be empty/not mapped in environments that don't support them as the Apply API does not return them).

- Sync additional pick list items for new CRM models

The new CRM models have additional pick list items (option sets) that need to be synced to ensure our validations work.


peteryates (approved):
Looks good

PR #964, opened 2022-04-22
Add ApplicationFormUpsert job

ethax-ross:

- Add generic Find/Apply CRM query

We are going to need to query application forms/choices/references/interviews from the CRM in order to reconcile the Apply API response with existing records based on the Find/Apply ID. Instead of having multiple CRM methods for each model we can leverage the `IHasFindApply` interface to make a generic method to query the models.

- Add job to upsert an ApplicationForm and related models

The `ApplicationForm` requires its own upserter job to match records to existing CRM models before persisting (as we used the `FindApplyId` rathern than the CRM `Id`).

Unlike the `Candidate` model it doesn't matter if an `ApplicationForm` gets part way through upserting and then fails; retrying will just update the partially created records from the previous job run and there's no issue with doing this (so we don't need to spawn a job per model we're updating).

It would be ideal if the CRM SDK we use supported deep-insert (which it used to but subsequently was found to be unreliable/flakey) and we could persist the entire form and related models in a single request, but for now this should
suffice and we can look into it if performance is an issue. Similarly, we make multiple requests to get existing models from the CRM whereas we _could_ load an application form and related models in a single request but this involves a more complicated query so I'm going to defer this until its needed.

- Rename UpsertModelJob to clarify intent

The `UpsertModelJob` actually deals with upserting a model related to a `Candidate` and will email the candidate if the job fails.

### DIFF
--- a/GetIntoTeachingApi/Jobs/FindApplyCandidateSyncJob.cs
+++ b/GetIntoTeachingApi/Jobs/FindApplyCandidateSyncJob.cs
@@ -102,7 +102,7 @@ namespace GetIntoTeachingApi.Jobs
 
             return findApplyApplicationForms.Select(findApplyForm =>
             {
-                var existingForm = _crm.GetApplicationForm(findApplyForm.Id.ToString(CultureInfo.CurrentCulture));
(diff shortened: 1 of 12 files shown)


peteryates (approved):
Looks good

PR #965, opened 2022-04-22
Support v1.2 of the Apply candidate API

ethax-ross:

- Ensure Apply API url is correct depending on features

If the new `APPLY_API_V2_1` feature is enabled we need to append `/v1.2` to the URL to ensure the correct response payload is returned from the Apply API.

- Use UpsertApplicationForm job when syncing Apply API

We have a new `UpsertApplicationFormJob` that deals with upserting either a v1.1 or v1.2 `ApplicationForm` and associated entities. Switching to use this in the `FindApplyCandidateSyncJob` simplifies the logic for updating (otherwise we would have a job per API version).


peteryates (approved):
👍🏽

PR #967, opened 2022-04-25
Add contract tests for the Apply API integration

ethax-ross:

- Extract contract test logic into base class

We are adding a second set of contract tests, so pulling all the code that can be shared from the existing `TeacherTrainingAdviserTests` into a `BaseTests` class.

- Add ApplyCandidateApi contract tests

We want to have contract tests covering the Apply API integration so that we can safely refactor/change the code and clearly know the knock-on effects in terms of data sent to the CRM.

Add `ApplyCandidateApiTests` to simulate both new and existing candidates/application forms.

Mock out the `ICrmService` so that we can stub requests for existing candidates/application forms to the CRM. Shift the `candidates.json` fixture so it can be shared across all contract tests and tweak it to be able to instantiate the models from JSON more easily.


peteryates (approved):
👍🏽

PR #969, opened 2022-04-26
Add contract tests for v1.2 of the Apply API integration

ethax-ross:
- Add contract tests for v1.2 of Apply API integration

Add additional contract tests to cover the new attributes/related models in the v1.2 Apply API integration.

- Clean up contract tests

Encapsulates more of the logic into the `BaseTests` superclass so that the contract tests are easier to understand.


peteryates (approved):
Nice

PR #970, opened 2022-04-26
Workflow to rebuild master docker image

NickGraham101:
### Context
Docker build layers are cached to optimise build performance but sometimes the base image is updated so we need to rebuild it.  Implementing snyk in this workflow highlights a high severity vulnerability in the base image.

- Add a build-no-cache workflow that is scheduled once a week on Sunday and can be run manually as required.
- Change the base image tag from the default to the alpine version.  This addresses the vulnerability, simplifies the dockerfile and aligns the OS with other projects which will simplify future maintenance. 

### Guidance to review
- Check the new workflow runs as expected, example [here](https://github.com/DFE-Digital/get-into-teaching-api/actions/runs/2232025759)
- Test the application to identify any issues with the OS change

### DIFF
--- a/.github/workflows/build-no-cache.yml
+++ b/.github/workflows/build-no-cache.yml
@@ -1,0 +1,74 @@
+name: Rebuild master docker image
+on:
+  workflow_dispatch:
+
+  schedule:
+    - cron: '0 12 * * 0'
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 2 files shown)


NickGraham101:
[Add workflow to bust docker cache](https://trello.com/c/HwLOMQ9D/546-add-workflow-to-bust-docker-cache)

ethax-ross (approved):

saliceti (approved):

PR #979, opened 2022-05-05
Increase HighJobsWaiting threshold

ethax-ross:
It looks like we periodically get a burst of enqueued jobs in SE; upping the limit to avoid false-positives.


peteryates (approved):
👍🏽

PR #989, opened 2022-05-13
Various dependency updates

ethax-ross:
The main update is to the `Dataverse.ServiceClient` which re-namepsaces the `PickListItem` model into `Extensions`. As we have our own `PickListItem` model we have to reference them explicitly via their namespace to avoid a conflict.

Closes #988, #987, #984, #968, #957


peteryates (approved):
Looks good

PR #995, opened 2022-05-20
Add new date fields to ApplicationReference

ethax-ross:
We want to be able to track the created/updated on dates for `ApplicationReference` entities; the Apply API has been updated to include these and this commit maps them into the respective CRM entity fields.


peteryates (approved):
Makes sense

PR #1038, opened 2022-08-05
Remove unused SecondaryEmail attribute from Candidate

ethax-ross:
[Trello-3515](https://trello.com/c/FF6qiIdL/3515-remove-emailaddress2-field-from-the-api)

We no longer collect a `SecondaryEmail` (we used to from the School Experience service). The CRM team have flagged that they are re-purposing that field and so we should remove it to make sure its not accidentally used in the future.


misaka (approved):
Er ... LGTM? 😛 Sorry, need to improve my C# skills.

PR #1044, opened 2022-08-12
[660] Update SE delayed job route

saliceti:
## What
https://github.com/DFE-Digital/schools-experience/pull/2617 removed a redundant route. It is used in the dev monitoring and should be updated to the correct one.

Also removed unused sandbox environment terraform config.


saliceti:
[SE delete review app: Error: The route could not be found](https://trello.com/c/KqqWtveD/660-se-delete-review-app-error-the-route-could-not-be-found)

ethax-ross (approved):

PR #1070, opened 2022-09-28
Re-instate stricter TeachingEvent.ReadableId validation

ethax-ross:
[Trello-3611](https://trello.com/c/2ib3KvKt/3611-reinstate-stricter-teaching-event-id-validation-on-provider-events)

We previously relaxed the validation around `ReadableId` due to some pre-existing events having unwanted characters in there; these events have since passed so we can make the validation stricter to ensure we don't get any more.


asatwal (approved):
👍  Looks Good to me .

PR #1081, opened 2022-10-06
Attempt to make the health check more robust

ethax-ross:
[Trello-656](https://trello.com/c/LivMzbVO/656-investigate-timeout-exception)

We are seeing errors on the School Experience service where the health check times out on calling the API health check endpoint.

I'm not sure what is causing this, but I expect that the CRM is being slow to respond as we've seen this happen in the past. To work around this we are only going to call out to the CRM once every minute (caching the health check responses in-between). 

I have also changed the CRM method we call to a generic entity name lookup for the `contact` entity; I suspect the other method we were using (`GetMyUserId`) may end up hitting Microsoft Authentication Library (MSAL) which could be a reason it takes longer than expected.

### DIFF
--- a/GetIntoTeachingApi/Adapters/OrganizationServiceAdapter.cs
+++ b/GetIntoTeachingApi/Adapters/OrganizationServiceAdapter.cs
@@ -23,7 +23,7 @@ namespace GetIntoTeachingApi.Adapters
         {
             try
             {
-                _client.GetMyUserId();
+                _client.GetEntityDisplayName("contact");
             }
             catch (Exception e)
             {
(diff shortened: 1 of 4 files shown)


GetIntoTeachingApi/Adapters/OrganizationServiceAdapter.cs
```diff
@@ -23,7 +23,7 @@ public string CheckStatus()
         {
             try
             {
-                _client.GetMyUserId();
+                _client.GetEntityDisplayName("contact");
```
asatwal, line 26: An obvious question :-) but does the CRM not have a healthcheck endpoint which is lightweight and just a basic ping ?

ethax-ross: If it does its sadly not available through the SDK we're using 😞 I checked their WebAPI docs and couldn't find anything there either so I don't think one exists/is publicly visible

asatwal (approved):
Looks good as far as I can tell.

PR #1088, opened 2022-10-17
[701] Pin github actions to major version

saliceti:
## What
Pin github actions to major version and let it pick the latest minor version automatically. This is to avoid creating many distracting dependabot PRs


saliceti:
[Remove github action minor version pinning](https://trello.com/c/BYSSVCNY/701-remove-github-action-minor-version-pinning)

PR #1089, opened 2022-10-18
Add a GSE alert for CSRF token errors

ethax-ross:
We want to be alerted if we experience a high number of CSRF token errors in a short period of time (indicating an underlying issue somewhere).


asatwal (approved):
👍  LGTM

PR #1090, opened 2022-10-19
Stop using set-output in workflows

smneal:
### Trello card
https://trello.com/c/3s3iC7UV

### Context
The set-output command is being deprecated soon.

Resolution is to use $GITHUB_OUTPUT as per the workflow commands docs.

### Guidance to review
Confirm updated workflows tested where possible:

- [ ] Deploy Other env (e.g. development)
- [ ] Build
- [ ] Build No Cache

### DIFF
--- a/.github/workflows/actions/deploy/action.yml
+++ b/.github/workflows/actions/deploy/action.yml
@@ -1,116 +1,116 @@
 name: Deploy to PaaS
 
 inputs:
-      environment:
-        description: The environment to deploy to Development/Test/Production/Review/Speed/UR etc
-        required: true
-      sha:
-        description: Commit sha to be deployed
-        required: true
-      pr:
-        description: Pull Request Reference
-        required: false
-      AZURE_CREDENTIALS:
-        required: true
-      KEY_VAULT:
-        required: true
-      ARM_ACCESS_KEY:
-        required: true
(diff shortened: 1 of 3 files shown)


.github/workflows/actions/deploy/action.yml
```diff
@@ -49,40 +49,40 @@ runs:
...
          id:  variables
          shell: bash
          run: |
              if [ "${{inputs.environment }}" == "Development" ]
              then
-                 echo ::set-output name=control::$(echo "dev" )
-                 echo ::set-output name=healthcheck::$(echo "${{env.PAAS_APPLICATION_NAME}}-dev" )
-                 echo ::set-output name=key::"api.dev.terraform"
-                 echo ::set-output name=docker_image::"${{env.DOCKER_REPOSITORY}}:sha-${{ steps.sha.outputs.short}}"
+                 echo "control=$(echo "dev" )" >> $GITHUB_OUTPUT
```
saliceti, line 60: We don't actually need `echo` here and in subsequent lines. Can we simplify?

smneal: Tried this but it failed. This is using the standard format in the documentation.

saliceti (approved):

PR #1091, opened 2022-10-20
Allow event search with radius but without postcode

ethax-ross:
[Trello-3784](https://trello.com/c/pzop8GSN/3784-address-500-errors-served-on-event-search)

The validation is a bit strict at the moment; if you provide a radius but no postcode we throw an error. The front-end currently allows this scenario and it makes sense to allow it from a users perspective as well.

Relax validation to return all results if filtering by radius without specifying a postcode (so treating it as if they haven't filtered by distance in this use case).


asatwal (approved):
👍  LGTM

PR #1093, opened 2022-10-25
Various dependency updates (inc major version bumps)

ethax-ross:
WIP

[Trello-3790](https://trello.com/c/hlRw9nZ7/3790-update-property-changed-fluent-validation)

Various dependency updates

```
- CsvHelper 28.0.1 -> 29.0.0
- Microsoft.EntityFrameworkCore 6.0.9 -> 6.0.10
- Microsoft.EntityFrameworkCore.Design 6.0.8 -> 6.0.10
- Microsoft.EntityFrameworkCore.Tools 6.0.8 -> 6.0.10
- Microsoft.VisualStudio.Web.CodeGeneration.Design 6.0.8 -> 6.0.10
- Npgsql.EntityFrameworkCore.PostgreSQL.NetTopologySuite 6.0.6 -> 6.0.7
- Sentry.AspNetCore 3.21.0 -> 3.22.0
- Serilog.Settings.Configuration 3.3.0 -> 3.4.0
- Microsoft.Extensions.Caching.StackExchangeRedis 6.0.9 -> 6.0.10
- YamlDotNet 12.0.1 -> 12.0.2
- GovukNotify 6.0.0 -> 6.1.0
- Microsoft.AspNetCore.DataProtection.StackExchangeRedis 6.0.8 -> 6.0.10
- Microsoft.Extensions.DependencyInjection 6.0.0 -> 6.0.1
- Serilog 2.11.0 -> 2.12.0
- Microsoft.EntityFrameworkCore.Relational 6.0.8 -> 6.0.10
- FluentAssertions 6.7.0 -> 6.8.0
- Microsoft.AspNetCore.Mvc.Testing 6.0.9 -> 6.0.10
- Microsoft.NET.Test.Sdk 12.3.1 -> 17.3.2
- WireMock.Net 1.5.6 -> 1.5.8
```

- Update Fody/PropertyChangedNotification

This is a major version bump; we have extensive tests around the functionality so it _should_ be safe, but I will do manual testing as well.

- Update Dataverse client

We use this to talk to the Dynamics CRM; its a patch up date but I'll do manual testing as well.

- Update FluentValidation

The `IValidationFactory` that we use to validate models coming back from the CRM has been deprecated and removed from the latest version of `FluentValidation`. I haven't found a nice way of doing this validation during reflection with the latest version; we have to inject the whole service provider into the models so that we can lookup the validator dynamically.

I may try refactoring this with a factory pattern in a future PR.

Also mocks out API calls in the rate limiting integration tests; I ended up doing this as part of chasing down a test failure which was actually caused by the validators not auto-registering with the controllers after the upgrade. The fix was to call `AddFluentValidationAutoValidation` during startup, but I'm leaving the mock in place as its more robust if we mock out the API (even though the validation prevents it being it in our current test suite I can see it being used in future tests).

### DIFF
--- a/GetIntoTeachingApi/AppStart/Startup.cs
+++ b/GetIntoTeachingApi/AppStart/Startup.cs
@@ -2,6 +2,7 @@
 using System.Text.Json.Serialization;
 using AspNetCoreRateLimit;
 using dotenv.net;
+using FluentValidation;
 using FluentValidation.AspNetCore;
 using GetIntoTeachingApi.JsonConverters;
 using GetIntoTeachingApi.ModelBinders;
@@ -46,8 +47,9 @@ namespace GetIntoTeachingApi.AppStart
             services.AddMvc(o => o.Conventions.Add(new CommaSeparatedQueryStringConvention()));
 
             services
+                .AddValidatorsFromAssemblyContaining<Startup>()
+                .AddFluentValidationAutoValidation()
                 .AddControllers(o => o.ModelBinderProviders.Insert(0, new TrimStringModelBinderProvider()))
-                .AddFluentValidation(c => c.RegisterValidatorsFromAssemblyContaining<Startup>())
                 .AddJsonOptions(o =>
                 {
                     o.JsonSerializerOptions.Converters.Add(new JsonStringEnumConverter());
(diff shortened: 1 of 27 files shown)


asatwal (commented):
👍 LGTM

PR #1094, opened 2022-10-26
Remove unused event search endpoint

ethax-ross:
The GiT website no longer calls the search events grouped by type endpoint (instead using the new search endpoint). It can be removed from the codebase (no other service uses it).


asatwal (approved):
👍  LGTM

PR #1114, opened 2022-11-30
Increase CdsSeviceClient timeout

ethax-ross:
We're seeing the current timeout of 15s being reached occasionally in production; increasing to 30s to see if it reduces the number of errors we get.


asatwal (approved):
👍

PR #1122, opened 2022-12-07
Reconcile equivalent email addresses

ethax-ross:
[Trello-4007](https://trello.com/c/Kmdw8ybc/3991-add-validation-in-api-to-treat-gmail-and-googlemail-as-the-same-address)

Some email poviders allow you to use multiple hosts that direct mail to the same mailbox. An example is `gmail.com` - if you have a GMail account you can also send/receive email from `googlemail.com`.

The CRM team have reported that the majority of duplicate contact records appear due to users signing up to a service with one domain and later using the other. Instead, they want to allow matchback on any variations that are equivalent (so point to the same mailbox/individual). This means a candidate can sign up for, say, the mailing list with `me@gmail.com` and then later register for an adviser using `me@googlemail.com` - the result will be a single contact record in the CRM.

For now we are only including GMail equivalent domains as they are the main contributor to duplicate records;  we may expand this in the future.


GetIntoTeachingApi/Utils/EmailReconciler.cs
```diff
@@ -0,0 +1,56 @@
...
+using YamlDotNet.Serialization;
+using YamlDotNet.Serialization.NamingConventions;
+using MoreLinq;
+using System.Linq;
+
+namespace GetIntoTeachingApi.Utils
+{
+	public static class EmailReconciler
+	{
+        private const string EquivalentEmailHostsFile = "./Fixtures/equivalent_email_hosts.yml";
```
asatwal, line 14: Is the `Fixtures` directory the place where config is normally kept with .net ? This sounds like a test fixture.

asatwal (approved):
LGTM. Just one question that I have.

GetIntoTeachingApi/Utils/EmailReconciler.cs
ethax-ross: Ah yeah I can see how that would be confusing; I'm not sure what the standard is for .net but I'll make a ticket to look at this and update it in a follow up PR

PR #1123, opened 2022-12-07
Ensure CRM email takes precedence over Apply

ethax-ross:
[Trello-4007](https://trello.com/c/U0JoQPO0/4007-dont-overwrite-contact-email-when-matching-on-apply-id)

When we matchback a candidate from the Apply service we use the Apply ID to ensure we get the correct CRM contact even if the email is different. Currently, we would overwrite the contact email address with the value from the Apply candidate, however going forward we want to give the CRM contact email precedence and retain it.

If the Apply candidate has a different email and there is no `SecondaryEmail` on the candidate we will write the Apply email to the `SecondaryEmail` field.


asatwal (approved):
👍 LGTM

PR #1124, opened 2022-12-07
Feature switch matchback on Apply ID

ethax-ross:
We don't want to matchback on Apply ID in production yet as there are still some open questions around which email we should take.

Wrap functionality in a feature switch and disable in production.

I noticed we're still using `launchSettings.json` to manage some local environment variables (such as feature switches). I've created a ticket to migrate these into the local dotenv, which is what we should be using going forward.


asatwal (approved):
👍  LGTM

PR #1127, opened 2022-12-09
Clean up local environment variables

ethax-ross:
[Trello-4008](https://trello.com/c/Lq1T6CgW/4008-migrate-local-env-vars-from-launchsettingsjson-to-envlocal)

Historically we used `launchSettings.json` to store non-sensitive, local environment variables. We have since introduced dotenv.net to manage our environment variables in `.env` files.

Migrate all local environment variables into `env.local`. Update the `Env` class to default the `CloudFoundryEnvironmentName` to `local` if the `VCAP_APPLICATION` environment variable is not set (this is a bit of a chicken/egg problem as the environment name is determined by the `VCAP_APPLICATION` environment variable, which isn't available until dotenv loads it).

The `ASPNETCORE_ENVIRONMENT` will remain in `launchSettings.json` as it needs to be set prior to the application booting.

### DIFF
--- a/GetIntoTeachingApi/Properties/launchSettings.json
+++ b/GetIntoTeachingApi/Properties/launchSettings.json
@@ -23,22 +23,7 @@
       "launchUrl": "swagger",
       "applicationUrl": "https://localhost:5001;http://localhost:5000",
       "environmentVariables": {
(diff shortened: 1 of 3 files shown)


asatwal (approved):
👍

PR #1130, opened 2022-12-15
Queue GSE sign up when CRM is offline

ethax-ross:
[Trello-643](https://trello.com/c/jl71Wn2O/643-implement-a-fallback-for-when-the-crm-is-sending-errors-when-gse-tries-to-write-a-record)

By default we create GSE candidates in-band as the service needs the created `contact.id` from the CRM in order to match up the GSE candidate with the CRM contact.

When the CRM integration is paused (this happens for maintenance occasionally) we don't want to prevent GSE sign ups, so we fallback to generating a `contact.id` ahead of persisting it in the CRM and we queue the upsert operation.

We had an incident a number of weeks ago where the CRM was offline unexpectedly for a period of time without being 'paused' (which the CRM team do manually prior to scheduled maintenance). During this time GSE sign ups were failing as the CRM could not be reached. Instead, we want to detect when the CRM is having issues and is _not_ in a paused state so that we can fallback to generating the `contact.id` ahead of time and queueing the upsert operation so that it can run when the CRM recovers.

The reason we don't generate the `contact.id` for all GSE sign up requests is that its better to rely on the auto-generated, sequential UUIDs that the CRM will create by default.

### DIFF
--- a/GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
+++ b/GetIntoTeachingApi/Controllers/SchoolsExperience/CandidatesController.cs
@@ -63,24 +63,26 @@ namespace GetIntoTeachingApi.Controllers.SchoolsExperience
                 return BadRequest(ModelState);
             }
 
+            // This is the only way we can mock/freeze the current date/time
+            // in contract tests (there's no other way to inject it into this class).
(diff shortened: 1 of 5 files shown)


asatwal (approved):
👍

PR #1132, opened 2022-12-20
Allow zero callback booking quotas to be returned

ethax-ross:
[Trello-4062](https://trello.com/c/n5iRbaIL/4062-review-callback-quota-fallback-functionality-in-light-of-increased-demand)

Currently we have fallback logic so that if the CRM is offline or 0 callback slots are available we are still going to return a default set of slots that will allow candidates to continue their sign up.

We recently had an unexpected influx of TTA sign ups that were being offered callback booking slots and found we had no way to stem the flow of bookings (if we mark all slots as available the default slots would then be offered). We currently offer 14 days in advance, so the likelihood of all slots being booked are slim, but we want to account for the scenario in case it does happen in the future.

Update the callback service to only return fallback slots when the CRM is offline. If the CRM is indicating there are no available slots we want to surface this on the websites and so we will now return 0 slots available.

### DIFF
--- a/GetIntoTeachingApi/Services/CallbackBookingService.cs
+++ b/GetIntoTeachingApi/Services/CallbackBookingService.cs
@@ -25,7 +25,7 @@ namespace GetIntoTeachingApi.Services
 
         public IEnumerable<CallbackBookingQuota> GetCallbackBookingQuotas()
         {
-            IEnumerable<CallbackBookingQuota> quotas = null;
+            IEnumerable<CallbackBookingQuota> quotas;
 
             try
             {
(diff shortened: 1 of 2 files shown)


asatwal (approved):
👍  LGTM.

Presumably the Sonar Cloud warnings were there before these changes and will be resolved sometime ?

ethax-ross:
> 👍 LGTM.
> 
> Presumably the Sonar Cloud warnings were there before these changes and will be resolved sometime ?

Yeah these warnings are unrelated, they're on my list but low priority

PR #1133, opened 2022-12-21
Make phone call/telephone optional for equivalent degrees

ethax-ross:
[Trello-4062](https://trello.com/c/n5iRbaIL/4062-review-callback-quota-fallback-functionality-in-light-of-increased-demand)

Currently, if a candidate has an equivalent degree we require them to schedule a callback with an adviser. Going forward, we are allowing these candidates to sign up without a callback in the event that there are no available callback booking slots available.

Trying to enforce the validation rules for when there are no callback slots available would be error prone, given that the availability dynamically updates. Instead, we are going to relax the validation in the API and ensure its enforced in the client services.


asatwal (commented):
👍 LGTM

asatwal (approved):

PR #1135, opened 2023-01-05
Disable Apply API integration

ethax-ross:
We are experiencing issues on the dev environment (its overloading the dev Apply candidate API). Disabling until we can fix it.


asatwal (approved):
👍

PR #1143, opened 2023-01-18
Revert "Disable Apply API integration"

ethax-ross:
Apply have updated their API so it should handle the traffic now.


asatwal (approved):
👍

PR #1145, opened 2023-01-24
Pin buildx to 0.9.1

ethax-ross:
New buildx generates a new OCI format with support for provenance. It doesn't work with Cloud foundry.


asatwal (approved):
👍

PR #1146, opened 2023-01-25
Add final year student contract test

ethax-ross:
We are treating final year students slightly differently so I want to capture this use case in a contract test so that we can be sure of what we're sending on to the CRM for this journey.


asatwal (approved):
👍  LGTM

PR #1147, opened 2023-01-26
Remove address 1/2 and city from TTA sign up

ethax-ross:
[Trello-4135](https://trello.com/c/J3a7zgtl/4135-remove-the-address-fields-from-the-adviser-funnel)

We no longer want to ask candidates for their full address during TTA sign up; instead we will only be asking for postcode going forward.

This relaxes the address validation for TTA sign ups but leaves in the other address fields so sign ups part way through will not be effected.


asatwal (approved):
👍  LGTM

PR #1148, opened 2023-01-27
Add API client for Apply service

ethax-ross:
The Apply service will be calling out to the API in order to sign up an Apply candidate for an adviser.

Add a client to the API configuration to support the Apply service. The relevant keys have already been added to the Azure keyvault for this client.


asatwal (approved):
👍  LGTM

PR #1149, opened 2023-01-30
Rename FindApply to Apply

ethax-ross:
[Trello-4041](https://trello.com/c/gOZOyWn1/4041-spike-submitting-tta-applications-from-the-apply-service)

When I initially integrated with the Apply service I thought Find/Apply was a single entity when in fact they're separate services. Renaming to clear up the naming.

Renames environment variables and features from `FindApplyApi` to `ApplyCandidateApi` (the secrets have been updated in keyvault accordingly).


asatwal (dismissed):
👍  LGTM

PR #1151, opened 2023-01-31
Updates to support adviser sign up from apply

ethax-ross:
[Trello-gOZOyWn1/4041](https://trello.com/c/gOZOyWn1/4041-spike-submitting-tta-applications-from-the-apply-service?filter=member:rossoliver15)

- Grant Apply access to adviser sign up

We are going to be accepting TTA sign ups from the Apply service going forward.

Grant access to the Apply service for TTA sign ups and candidate information.

- Add endpoint to retrieve existing candidate id

When a candidate signs up for a TTA via the Apply service they will already be 'authenticated' against the candidate email so we will not need to issue them a TOTP. Instead, we want to obtain a candidate id by providing matchback information.

- Setup rate limits for Apply service

Add rate limits for the adviser sign up from Apply service and also the new matchback endpoint as we're hitting the CRM.

### DIFF
--- a/GetIntoTeachingApi/Controllers/CandidatesController.cs
+++ b/GetIntoTeachingApi/Controllers/CandidatesController.cs
@@ -13,7 +13,7 @@ namespace GetIntoTeachingApi.Controllers
 {
     [Route("api/candidates")]
     [ApiController]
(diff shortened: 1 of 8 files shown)


asatwal (approved):
👍 LGTM

PR #1153, opened 2023-02-13
Add adviser matchback endpoint for Apply service

ethax-ross:
[Trello-4041](https://trello.com/c/gOZOyWn1/4041-spike-submitting-tta-applications-from-the-apply-service)

The Apply service is unlike our other API clients in that it has a set of candidates details in its own database (whereas other clients rely on the API/CRM for candidate information). As the Apply candidates are pre-verified by the Apply service we aren't going to require them to enter a Timed One-Time Passcode to retrieve existing candidate information from the API. Instead, the service can issue a matchback request for existing candidate details using the standard matchback rules.

Remove the generic candidate matchback endpoint as this doesn't fit in with the user journey we envisage on Apply for an adviser sign up (we need to retrieve adviser-specific details not just the existing candidate ID).


asatwal (approved):
👍  LGTM

PR #1160, opened 2023-02-14
Relax GCSE science validation

ethax-ross:
The Apply service will allow candidates to sign up with Maths/English GCSEs (and no Science GCSE) regardless of their preferred education stage, so we can relax this validation rule.


asatwal (approved):
👍

PR #1172, opened 2023-02-22
Address code climate warnings

ethax-ross:
- Set global Regex timeout to 500ms to prevent potential DoS attack
- Swap `ToList` for `AsEnumerable`
- Remove redundant code


asatwal (approved):
👍 LGTM

PR #1175, opened 2023-02-23
Remove redundant array

ethax-ross:
Flagged by CodeClimate; the array here is redundant and we can just pass the labels directly.

### DIFF
--- a/GetIntoTeachingApi/Jobs/UpsertModelWithCandidateIdJob.cs
+++ b/GetIntoTeachingApi/Jobs/UpsertModelWithCandidateIdJob.cs
@@ -78,7 +78,7 @@ namespace GetIntoTeachingApi.Jobs
             }
 
             var duration = (DateTime.UtcNow - _contextAdapter.GetJobCreatedAt(context)).TotalSeconds;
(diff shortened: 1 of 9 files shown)


GetIntoTeachingApi/Jobs/UpsertModelWithCandidateIdJob.cs
```diff
@@ -78,7 +78,7 @@ public void Run(string json, PerformContext context)
             }
 
             var duration = (DateTime.UtcNow - _contextAdapter.GetJobCreatedAt(context)).TotalSeconds;
-            _metrics.HangfireJobQueueDuration.WithLabels(new[] { $"UpsertModelJob<{typeName}>" }).Observe(duration);
+            _metrics.HangfireJobQueueDuration.WithLabels($"UpsertModelJob<{typeName}>").Observe(duration);
```
asatwal, line 81: is the `$` required in the argument ?

asatwal (approved):
👍 LGTM - Just one question that I have.

GetIntoTeachingApi/Jobs/UpsertModelWithCandidateIdJob.cs
ethax-ross: Yeah its for interpolating the `{typeName}` into the string

PR #1177, opened 2023-02-27
Remove LookupItem class

ethax-ross:
[Trello-4200](https://trello.com/c/hKLgzDqy/4200-remove-lookupitem-class)

We no longer reference this class as it was replaced by `Country` and `TeachingSubject`. We kept it in place so there was no downtime during deployment of the replacement classes, but it can now be safely removed.


asatwal (approved):
👍 LGTM

PR #1178, opened 2023-02-27
Remove unused address fields from adviser sign up

ethax-ross:
[Trello-4151](https://trello.com/c/cYOL1Few/4151-clean-up-adviser-funnel-address-postcode-fields)

We no longer ask for the address line 1/2 and city as part of the adviser sign up; these fields can be removed.


asatwal (approved):
👍 LGTM

PR #1182, opened 2023-03-06
Matchback on secondary email address

ethax-ross:
[Trello-4006](https://trello.com/c/ytNAZcUw/4006-check-secondary-email-field-on-matchback)

Currently we only matchback candidates against the primary email of the CRM contact. Going forward we want to matchback on either the primary or secondary email addresses.


asatwal (approved):
👍 LGTM

PR #1185, opened 2023-03-08
Rename fixtures directory to config

ethax-ross:
[Trello-4309](https://trello.com/c/GFOv3Fmk/4309-rename-fixtures-directory-in-api-to-be-clearer)

The term 'fixtures' is often associated with test fixtures which makes this a bit confusing; renaming to 'config' to make it clearer.


asatwal (approved):
👍

PR #1186, opened 2023-03-08
Clean up BaseModel hooks

ethax-ross:
[Trello-4304](https://trello.com/c/gjjriiec/4304-clean-up-basemodel-hooks-shift-into-candidateupserter)

- Remove ShouldMapRelationship hook

The `BaseModel` is complex and more so due to the available hooks. We have since centralised the candidate updating to the `CandidateUpserter` which makes the hooks redundant. Port the privacy policy hook over to the candidate upserter.

- Remove ShouldMap hook from BaseModel

The logic currently in the `ShouldMap` hooks can be moved into the `CandidateUpserter` to avoid overcomplicating an already complex class.

The only hook remaining in `BaseModel` is the `FinaliseEntity` hook, which can't be removed as it operates on the encapsulated CRM entity.


asatwal (approved):
👍 LGTM

PR #1194, opened 2023-03-13
Expose adviser assignment status

ethax-ross:
We want to show a different UI to candidates in Apply based on their adviser assignment status; in order to do that we need to expose it from the API.


asatwal (approved):
:+1

PR #1196, opened 2023-03-17
Relax ExistingCandidateRequest validation

ethax-ross:
[Trello-4360](https://trello.com/c/IfbKWsDm/4360-investigate-400-error-on-adviser-sign-up-from-apply) 

Historically the matchback was more stringent and required a first/last name or DOB to match as well as email. It has since been updated to fallback to only matching on email address, although all existing services will also send a first/last name.

Now that Apply offers an adviser sign up the matchback can happen without a first/last name and the API raises an error. Relaxing this validation will prevent the error and not change the underlying functionality (as we fallback to an email only matchback anyway).


asatwal (approved):
👍  LGTM

PR #1200, opened 2023-03-22
Update adviser assignment logic

ethax-ross:
[Trello-4363](https://trello.com/c/Tg9tgd58/4363-investigate-apply-adviser-sign-up-issues)

Currently, if a candidate has a degree or is returning to teaching we will put them in the queue to be assigned to an adviser. Other candidates (with an equivalent degree) are taken on a journey to schedule a callback on the main TTA service.

If an international candidate signs up from Apply we do not offer the callback service and they aren't put in the queue for an adviser, so they will fall between the cracks. Going forward we want to queue these candidates for an adviser. To that end, the logic around this can be simplified to 'if a candidate has not booked a callback, put them in the queue for an adviser'.


asatwal (approved):
👍

PR #1201, opened 2023-03-23
Disable ApplySyncJob in production

ethax-ross:
We are planning to backfill the Apply candidate information into the production CRM this afternoon; before we do that the CRM team will clear down existing production data and have requested we pause the hourly sync job.


asatwal (approved):
👍  LGTM

PR #1203, opened 2023-03-23
Remove feature flags

ethax-ross:
[Trello-4366](https://trello.com/c/VGMplzV4/4366-remove-feature-switching-around-matchback-and-apply-api-sync)

These feature flags are now enabled in all environments and are safe to remove. Also pins the apply endpoint version in all environments to `1.2`.


asatwal (approved):
👍 LGTM

PR #1208, opened 2023-03-24
Consolidate matchbacks for consistency

ethax-ross:
[Trello-4371](https://trello.com/c/RlYbE0Ec/4371-fix-matchback-issue)

Currently we matchback slightly differently for an email-only matchback (as used by the Apply candidate sync jobs) and a matchback that uses the `ExistingCandidateRequest` model. Going forward we want to only matchback on email (ignoring first/last/date of birth); the `ExistingCandidateRequest` matchbacks were performing further checks and an additional email comparison that was overwriting the `EmailReconciler` logic by doing an explicit equality check.

The `ExistingCandidateRequest` is retaining the additional attributes for now so that the APIs don't change, but going forward we can pair this class back to only having an `Email` attribute.


asatwal (approved):
👍

PR #1210, opened 2023-03-27
Re-enable Apply sync in production

ethax-ross:
Now that we have completed the backfill we can re-enable the apply data sync in production to get hourly updates.


asatwal (approved):
👍

PR #1211, opened 2023-03-27
Enable v2 matchback rules in production

ethax-ross:
These have been tested by the CRM team now and can be enabled in production.


asatwal (approved):
👍

PR #1214, opened 2023-03-29
Disable Hangfire dashboard in production

ethax-ross:
[Trello-4412](https://trello.com/c/MxAmntgn/4412-disable-hangfire-dashboard-in-production)

We enabled this (behind basic auth) to more easily monitor the Apply backfill job. Re-disabling it now that's finished.


asatwal (approved):
👍

PR #1215, opened 2023-03-30
Revert "Disable Hanfire Prometheus metrics exporter"

ethax-ross:
Reverts DFE-Digital/get-into-teaching-api#1209

I think this was just an issue due to the unusually high volumes of jobs we queued up during the recent Apply backfill operation; reverting it so we get job metrics back in production.


asatwal (approved):
👍

PR #1217, opened 2023-03-31
Require first name on matchback

ethax-ross:
[Trello-4372](https://trello.com/c/iDqBrbp4/4372-clean-up-existingcandidaterequest-model)

We only use the email address in the matchback logic itself, but we reference the candidate by their first name in the email that we send out with their access code; we need to retain this field/ensure make it mandatory.


asatwal (approved):
👍

PR #1225, opened 2023-04-12
Grant GetIntoTeaching role access to adviser sign up

ethax-ross:
[Trello-4444](https://trello.com/c/0PFn0Tb7/4444-spike-investigate-moving-the-adviser-funnel-to-git)

We are planning on migrating the TTA service into the GiT website. For the spike the GiT app needs to be able to submit to the TTA endpoints; once we're happy with the solution the API namespaces will be updated to reflect the TTA service migration into GiT.


asatwal (approved):
👍

PR #1227, opened 2023-04-17
Fix metrics erroring out after update

ethax-ross:
[Trello-4472](https://trello.com/c/lrqFaQTM/4472-fix-grafana-panels-re-one-time-passwords)

We recently updated our .NET Core Prometheus Client and the update appears to have introduced a bug with exemplars. Since we don't use exemplars at the moment we can disable the behaviour as a workaround until its patched in the library.


asatwal (approved):
👍

PR #1265, opened 2023-07-05
Add env var to check redis connection string is present

plisovin:


divesh987:
[Remove VCAP services dependencies](https://trello.com/c/jJUicruz/465-remove-vcap-services-dependencies)

PR #1268, opened 2023-07-11
Fix redis configuration in case redis connection string is used for aks

plisovin:


divesh987 (approved):

divesh987:
[Remove VCAP services dependencies](https://trello.com/c/jJUicruz/465-remove-vcap-services-dependencies)

PR #1321, opened 2023-11-16
[669] Update the buildx action

saliceti:
## What
It was pinned to 0.9.1 for compatibility with GOV.UK PaaS. Now that we run on AKS, this limitations does not apply and we can use the default version.


saliceti:
[GIT-API: Decommission paas environments and code](https://trello.com/c/AKvLo2fS/669-git-api-decommission-paas-environments-and-code)

johnake (approved):

PR #1329, opened 2023-12-11
[669] Clean up code after migration

saliceti:
## Description
- Deleted paas deploment code
- Removed references to paas
- Reintroduced missing smoke test
- Fixed labeler config for V5


.github/labeler.yml
```diff
@@ -1,18 +1,23 @@
 ---
 
 VisualCSharp:
-  - GetIntoTeachingApi/**/*.cs
+- changed-files:
```
martyn-w, line 4: Thanks for updating this - however, the labeler is still throwing errors:

```
Run actions/labeler@main
The configuration file (path: .github/labeler.yml) was not found locally, fetching via the api
Error: Error: found unexpected type for label 'VisualCSharp' (should be array of config options)
Error: found unexpected type for label 'VisualCSharp' (should be array of config options)
```

saliceti: Yes it's a bit confusing but it turns out the "pull_request_target" trigger picks up the yml file from the master branch, so it will only work after merging.
I did test it with the "pull_request" trigger which picked up the yml file from the PR branch and it worked: https://github.com/DFE-Digital/get-into-teaching-api/actions/runs/7183796826

martyn-w (approved):
👍  Looks good - thanks very much for tidying this up 👍

.github/labeler.yml
martyn-w: Jolly good - that would explain it!

PR #1342, opened 2024-01-19
high availability disabled

VenkataChalla554:
SKU and High availability upgrades can't be done same time. Disable High availability .


saliceti:
Not required anymore, closing

PR #1433, opened 2024-06-06
Bugfix to apply candidate sync job

martyn-w:
This corrects a bug introduced by the upgrade of the `Flurl.Http` library in PR #1383 which changed the default json parsing library from `Newtonsoft.Json` to `System.Text.Json`. A consequence of this bug was that parsing of the Apply Candidates API fails and so synchronisation between Apply and the CRM failed.

This fix imports the library `Flurl.Http.Newtonsoft` and adds calls to `FlurlHttp.Clients.UseNewtonsoft()` to enforce usage of the Newtonsoft serializer.

For more information, see:

* https://flurl.dev/docs/upgrade/
* https://www.nuget.org/packages/Flurl.Http.Newtonsoft/#readme-body-tab


asatwal (approved):
👍 LGTM

PR #1464, opened 2024-08-15
New endpoint for synchronising by candidate ID

martyn-w:
Adds a new endpoint, `/api/operations/backfill_apply_candidates_from_ids` for triggering a synchronise with the Apply system for a specific Candidate ID.

See trello ticket: https://trello.com/c/xcork9SB/6409-backfill-missing-data-from-apply-into-the-crm

### DIFF
--- a/GetIntoTeachingApi/Controllers/OperationsController.cs
+++ b/GetIntoTeachingApi/Controllers/OperationsController.cs
@@ -142,8 +142,31 @@ namespace GetIntoTeachingApi.Controllers
             {
                 return BadRequest("Backfill already in progress");
             }
+            
+            _jobClient.Enqueue<ApplyBackfillJob>((x) => x.RunAsync(updatedSince, 1, null));
 
-            _jobClient.Enqueue<ApplyBackfillJob>((x) => x.RunAsync(updatedSince, 1));
+            return NoContent();
+        }
+        
+        [HttpPost]
+        [Authorize(Roles = "Admin")]
+        [Route("backfill_apply_candidates_from_ids")]
+        [SwaggerOperation(
+            Summary = "Triggers a backfill job to sync the CRM with the Apply candidates for specified candidate Ids.",
+            Description = "The backfill will query all candidate information from the Apply API and " +
+                          "queue jobs to sync the data with the CRM.",
+            OperationId = "BackfillApplyCandidatesFromIds",
+            Tags = new[] { "Operations" })]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        public IActionResult BackfillApplyCandidatesFromIds([FromBody, SwaggerRequestBody("Candidate IDs to backfill", Required = true)] CandidateIdsRequest request)
+        {
+            if (_appSettings.IsApplyBackfillInProgress)
+            {
+                return BadRequest("Backfill already in progress");
+            }
+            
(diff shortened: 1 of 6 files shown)


GetIntoTeachingApi/Controllers/OperationsController.cs
```diff
@@ -2,13 +2,13 @@
...
-using GetIntoTeachingApi.Jobs;
+using GetIntoTeachingApi.Jobs;
 using GetIntoTeachingApi.Models;
 using GetIntoTeachingApi.Models.Crm;
 using GetIntoTeachingApi.Services;
 using GetIntoTeachingApi.Utils;
-using Hangfire;
-using Microsoft.AspNetCore.Authorization;
+using Hangfire;
+using Microsoft.AspNetCore.Authorization;
```
carlosmartinez, line 11: why do these show up as diffs?

GetIntoTeachingApi/Jobs/ApplyBackfillJob.cs
```diff
@@ -76,7 +88,48 @@ private async Task QueueCandidateSyncJobs(DateTime updatedSince, int startPage)
...
+                {
+                    _logger.LogError("Failed to fetch CandidateID C{CandidateId} from the Apply API (status: {Status})", candidateId, ex.StatusCode);
+                }
+
+                await Task.Delay(100);
+            }
+            
+            // When we reach the end page we re-queue the backfill job
+            // to process the next batch of candidate IDs.
+            if (remainder?.Any() ?? false)
```
carlosmartinez, line 130:
```suggestion
            if (remainder?.Any())
```

GetIntoTeachingApi/Jobs/ApplyBackfillJob.cs
```diff
@@ -76,7 +88,48 @@ private async Task QueueCandidateSyncJobs(DateTime updatedSince, int startPage)
...
+                    _logger.LogInformation("Scheduling ApplyBackfillJob - Syncing CandidateID: C{Id}", candidateId);
+                    
+                    _jobClient.Schedule<ApplyCandidateSyncJob>(x => x.Run(candidate.Data), TimeSpan.FromSeconds(60));
+                }
+                catch (FlurlHttpException ex)
+                {
+                    _logger.LogError("Failed to fetch CandidateID C{CandidateId} from the Apply API (status: {Status})", candidateId, ex.StatusCode);
+                }
+
+                await Task.Delay(100);
```
carlosmartinez, line 125: is this for rate limiting? probably deserves a comment if so

GetIntoTeachingApi/Jobs/ApplyBackfillJob.cs
carlosmartinez: i'm guessing `null` if falsey in C#

carlosmartinez (dismissed):

GetIntoTeachingApi/Jobs/ApplyBackfillJob.cs
martyn-w: Sadly not, a fun fact about C# - `null` does not cast to a boolean:
```
public static void Main()
{
	if (null) {
		Console.WriteLine("Hello World");
	}
}
```

is a compilation error: `Compilation error (line 7, col 7): Cannot convert null to 'bool' because it is a non-nullable value type`


instead, you can use the double question mark operator `??` to return false if the previous arg was a null, e.g.

```
some_value_which_might_be_null ?? false
```

...of course, Ruby solved this problem 30 years ago...

carlosmartinez: dear lord...

carlosmartinez (approved):

PR #1467, opened 2024-09-05
Prevent duplicate degree subject

martyn-w:
This change improves the synchronisation of qualification records:

* adds checks to ensure a qualification record is not re-created if it already exists
* creates qualification records immediately, rather than via a queued job

### DIFF
--- a/GetIntoTeachingApi/Models/SchoolsExperience/SchoolsExperienceSignUp.cs
+++ b/GetIntoTeachingApi/Models/SchoolsExperience/SchoolsExperienceSignUp.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using System.Text.Json.Serialization;
 using GetIntoTeachingApi.Models.Crm;
 using GetIntoTeachingApi.Services;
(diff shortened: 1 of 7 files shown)


GetIntoTeachingApi/Models/SchoolsExperience/SchoolsExperienceSignUp.cs
```diff
@@ -132,5 +150,25 @@ private void AcceptPrivacyPolicy(Candidate candidate)
...
+                    DegreeStatusId = DegreeStatusId,
+                    DegreeSubject = DegreeSubject,
+                    TypeId = DegreeTypeId ?? (int)CandidateQualification.DegreeType.Degree
+                });
+            }
+        }
+        
+        private bool ContainsQualification()
+        {
+            return UkDegreeGradeId != null || DegreeStatusId != null || DegreeSubject != null || DegreeTypeId != null;
```
asatwal, line 171: Are any of these fields mandatory for a Degree to be valid ? Vould missing fiedls here cause problems with the downstream systems ?

asatwal (commented):
Just one question below.

GetIntoTeachingApi/Models/SchoolsExperience/SchoolsExperienceSignUp.cs
martyn-w: nope, they are intentionally optional as older clients will not send the new degree fields. Missing these values means that a qualification record will not be created in the CRM (as is currently the case), but shouldn't break anything downstream.

asatwal (approved):
👍 LGTM

PR #1469, opened 2024-09-17
Remove request logging from API

martyn-w:
Remove request logging from the API - requested by infra team to reduce logging volumes.


GetIntoTeachingApi/Middleware/RequestResponseLoggingMiddleware.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System.IO;
 using System.Linq;
-using System.Threading.Tasks;
+using System.Threading.Tasks;
```
carlosmartinez, line 3: worth looking at the git config to avoid seeing whitespace changes in the diff?

carlosmartinez (approved):

PR #1495, opened 2025-01-14
Add mailing list creation channel

martyn-w:
Adds Creation Channel to the mailing list end point

https://trello.com/c/iVQSw2yD/7071-update-git-api-to-support-new-creation-channel-structure-for-subscription-data-e11-gc3

### DIFF
--- a/GetIntoTeachingApi/Models/GetIntoTeaching/MailingListAddMember.cs
+++ b/GetIntoTeachingApi/Models/GetIntoTeaching/MailingListAddMember.cs
@@ -20,6 +20,13 @@ namespace GetIntoTeachingApi.Models.GetIntoTeaching
         public int? DegreeStatusId { get; set; }
         [SwaggerSchema(WriteOnly = true)]
         public int? ChannelId { get; set; }
+        
+        [SwaggerSchema(WriteOnly = true)]
+        public int? CreationChannelSourceId { get; set; }
+        [SwaggerSchema(WriteOnly = true)]
+        public int? CreationChannelServiceId { get; set; }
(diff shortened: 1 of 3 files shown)


GetIntoTeachingApi/Models/GetIntoTeaching/MailingListAddMember.cs
```diff
@@ -104,9 +111,38 @@ private void ConfigureChannel(Candidate candidate)
         {
```
spanersoraferty, line 111: Just wondering if we can flatten this logic a bit...

something like (although not sure if this logic is a 100% match) ...

 **bool creationChannel =
     CandidateId == null && !candidate.ContactChannelCreations.Any();

 if (!creationChannel) { candidate.ChannelId = null; }

 AddCandidateCreationChannel(candidate, creationChannel);**

martyn-w: Hi @spanersoraferty, unfortunately that suggested logic does not match the desired behaviour.

Actually, as the logic is a little bit fiddly, and I am inclined to leave it in its current form so that it can be easily(ish) traced.

spanersoraferty: Hi @martyn-w , I wasn't super confident with the logic, it was more a steer on maybe how to flatten the conditional statements (I think I spotted the issue with my quick attempt BTW :) BUT I totally understand your hesitation in wanting to make any changes so happy to approve.

spanersoraferty (approved):

PR #1498, opened 2025-01-28
So/review sentry exception logging in the api

spanersoraferty:
I've limited the use of the correlation ID to the teacher advisor sign-up endpoint (as discussed) to limit the potential for issues if any problems occur. I've also not been able to best the filter attribute, it is possible but I cannot mock the hangfire context because they've neglected to apply interfaces, or make class methods generic. We could use a library like Typemock but not sure if this will play nicely with the CICD pipeline out of the box.

If theses changes prove stable we can raise a second, follow-on ticket to complete the work to apply this logging behaviour across the remaining endpoints.


GetIntoTeachingApi/Jobs/UpsertCandidateJob.cs
```diff
@@ -38,23 +39,26 @@ public UpsertCandidateJob(
...
             if (_appSettings.IsCrmIntegrationPaused)
             {
-                throw new InvalidOperationException("UpsertCandidateJob - Aborting (CRM integration paused).");
+                throw new InvalidOperationException(
+                    $"UpsertCandidateJob - Aborting (CRM integration paused).");
             }
 
-            _logger.LogInformation("UpsertCandidateJob - Started ({Attempt})", AttemptInfo(context, _contextAdapter));
-            _logger.LogInformation("UpsertCandidateJob - Payload {Payload}", Redactor.RedactJson(json));
+            _logger.LogInformation("UpsertCandidateJob - Started ({Attempt}) {CorrelationId}", AttemptInfo(context, _contextAdapter), correlationId);
```
martyn-w, line 60: could the correlation Id be made more prominent in the log so it is easy to distinguish from all the other GUIDs? How about:
* `CID-162cdb97-32ba-4efc-b7f8-7a2ad87b8d39`
* `[ 162cdb97-32ba-4efc-b7f8-7a2ad87b8d39]`
* `( 162cdb97-32ba-4efc-b7f8-7a2ad87b8d39)`

GetIntoTeachingApi/appsettings.json
```diff
@@ -10,9 +10,18 @@
     "WriteTo": [
       {
         "Name": "Console"
+      },
+      {
+        "Name": "File",
```
martyn-w, line 15: could this be a local dev setting, rather than in the main app settings? I don't think this will work on production

spanersoraferty: Actually, this should be removed. Thought I had, leave it with me and I'll sort.

GetIntoTeachingApi/Jobs/UpsertCandidateJob.cs
spanersoraferty: No problem, happy to do whatever you think may help. Leave it with me and I'll send you the output once I've applied the change to see if things look right.

martyn-w: looks good 👍

martyn-w (approved):
👍 LGTM, tested locally 👍

PR #1499, opened 2025-02-05
So/7343 update .net version in git api

spanersoraferty:
This has been a little bit more involved than I had originally hoped, and has required some code changes to facilitate the upgrade but all the tests appear happy. I've also update the NuGet dependencies where required but tried to keep as lighter touch as possible here as well.


martyn-w (approved):
👍 LGTM

PR #1522, opened 2025-04-14
Restructured and added more narrative around the background processin…

spanersoraferty:
Some restructuring and first couple of sections around background processing and database details extended.


martyn-w (approved):
👍 LGTM

PR #1535, opened 2025-05-01
Removal of unused dependencies in API

spanersoraferty:
Removed some third part NuGet package dependencies that are not being used.


martyn-w (approved):
👍  LGTM

PR #1547, opened 2025-05-27
Initial commit - added event accessibility picklist endpoint

spanersoraferty:


martyn-w (changes requested):
Hi Spencer, I've tried to run this locally but I don't think this working yet.

* I think you'll need to add a line to `GetIntoTeachingApi/Services/Store.cs` to synchronise the new accessibility statuses
* Could you use a pluralised name for the endpoint, to match existing endpoints
* Could you check the Narendra the actual field name and confirm there is data in the table?

thanks!

PR #1576, opened 2025-09-03
Adding in new dfe_accessibilityneedsforevent

Edu-GG:
Adding in new field to allow attendees to update accessibility needs for events


GetIntoTeachingApi/Models/GetIntoTeaching/TeachingEventAddAttendee.cs
```diff
@@ -165,7 +173,7 @@ private Candidate CreateCandidate()
             AddQualification(candidate);
             AcceptPrivacyPolicy(candidate);
             ConfigureSubscriptions(candidate);
-
+
```
martyn-w, line 176: discard this blank line

PR #1581, opened 2025-09-16
Gg/8231 api implement dfeukdegreegrade field to support adviser eligibility triage

Edu-GG:
Adding additional Validation for ukDegreeGrade.
Accepting only valid listed options on submission.
Test modified to match valid options.

### DIFF
--- a/GetIntoTeachingApi/Models/Crm/CandidateQualification.cs
+++ b/GetIntoTeachingApi/Models/Crm/CandidateQualification.cs
@@ -14,6 +14,16 @@ namespace GetIntoTeachingApi.Models.Crm
             Degree = 222750000,
             DegreeEquivalent = 222750005,
         }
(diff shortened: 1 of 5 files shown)


GetIntoTeachingApi/Models/Crm/CandidateQualification.cs
```diff
@@ -14,6 +14,16 @@ public enum DegreeType
             Degree = 222750000,
             DegreeEquivalent = 222750005,
         }
+        
+         public enum UkDegreeGrade
+        {
+            LowerSecond = 222750003,
```
martyn-w, line 20: could you move this to line 24 so they are ordered by ID - it will help in future if we need to add new items

martyn-w (approved):
👍 LGTM

PR #1583, opened 2025-09-22
Add non root for git api container

shaheislamdfe:


Edu-GG (approved):
Tested ok in review app

PR #1584, opened 2025-10-02
dropping validation on path

Edu-GG:
remove some validation to take into account path for non uk degree


carlosmartinez (approved):
seems reasonable!

PR #1640, opened 2026-06-22
Run PR workflow for docs changes

martyn-w:
Remove 'docs/**' from paths-ignore in .github/workflows/pull_request.yml so the pull_request workflow will run when documentation files are modified. Ensures CI checks are executed for doc updates.


hopep-mobilise (approved):
Blocking docs is no longer needed

PR #1642, opened 2026-06-23
2683 - Missing with var and removing domains

hopep-mobilise:
## Context

Update the validate infra workflow.

https://trello.com/c/p76gfyrn/2683-schedule-validate-infra-workflow

## Changes proposed in this pull request

Add the variable definition for terraform-version-file in the with of steps.
Remove domains as they are not used but were disguised under vendor.

## Guidance to review

Run the "Validate Infra" workflow as it already exists for the service. You can point it at the branch but that will always produce a false positive as the docker image will be different for main and the branch but you should see that the structure is correct and it's just the sha that's different. This will produce a workflow Teams message in SD Infra Alerts Teams channel.

image

There are 3 jobs in the workflow, AKS infrastructure, Domains infrastructure and Domains environment. The workflow card will show which job to look at in the bottom left corner (Infrastructure in the image above). You can click the Workflow Run button to go to the correct run.
The jobs must show a green tick not a red cross.
You can alter any value in the terraform production.tfvars.json in your branch to check that the terraform plan picks up those changes, remembering that the docker image is a false positive. 

## Checklist

- [x] I have performed a self-review of my code, including formatting and typos
- [x] I have [cleaned the commit history](https://www.annashipman.co.uk/jfdi/good-pull-requests.html)
- [x] I have added the `Devops` label
- [x] I have attached the pull request to the trello card

### DIFF
--- a/.github/workflows/validate-infrastructure.yml
+++ b/.github/workflows/validate-infrastructure.yml
@@ -21,12 +21,6 @@ jobs:
             validation_plan: aks-infra
             terraform_base: terraform/aks
             terraform_version_file: provider.tf
-          - display: Domains infrastructure
-            validation_plan: domains-infra
-            terraform_base: terraform/domains/infrastructure
-          - display: Domains environment
-            validation_plan: domains-env
-            terraform_base: terraform/domains/environment_domains
 
     steps:
       - name: Validate ${{ matrix.display }}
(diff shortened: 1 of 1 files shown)


matt-mobilise (approved):
Checked domains not used and code looks fine to me

PR #1644, opened 2026-06-24
Remove proposed change from /api/schools_experience/candidates

CatalinVoineag:
After some consideration. We need to remove the proposed change for this POST endpoint /api/schools_experience/candidates.

The client using this endpoint relies on a live response. We cannot enqueue the creation of a candidate in a background job.


martyn-w (approved):
👍 LGTM

PR #1649, opened 2026-07-22
CRM api spec V2

CatalinVoineag:
Following feedback from people, I've updated the CRM api spec with the endpoints that are needed and added some proposed changes to other endpoints.

Mainly the endpoints that deal with tokens. They will just act as GET endpoints that search for a record to check if they exsits. Generating tokens and validating them will not be the responsibility of the CRM.

This also removes a few endpoints. Backfill/Operation endpoints are not needed. The backfill will be done separately. Some /attendees endpoints are not needed anymore.

Viewing the changes with a markdown viewer is easier


dcyoung-dev (dismissed):

martyn-w (approved):
👍 LGTM

